PR #3, opened 2022-09-15
Elex-1304-interval fix

daragold:
In making aggregate prediction intervals on non-reporting units in the Gaussian model, the minimum allowed was 0, instead of the votes already returned in those units. Changed this minimum in GaussianElectionModel.py lines 227/228


src/elexmodel/models/BaseElectionModel.py
```diff
@@ -111,6 +111,15 @@ def _get_reporting_aggregate_votes(self, reporting_units, unexpected_units, aggr
 
         return aggregate_votes
 
+    def _get_nonreporting_aggregate_votes(self, nonreporting_units, unexpected_units, aggregate, estimand):
```
lennybronner, line 114: please add a unit test for this function

src/elexmodel/models/GaussianElectionModel.py
```diff
@@ -220,8 +224,8 @@ def get_aggregate_prediction_intervals(
         aggregate_prediction_intervals = (
             last_election.merge(modeled_bounds, how="inner", on=aggregate)
             .assign(
-                predicted_lower=lambda x: np.maximum(x[f"last_election_results_{estimand}"] + x.lb, 0),
-                predicted_upper=lambda x: np.maximum(x[f"last_election_results_{estimand}"] + x.ub, 0),
+                predicted_lower=lambda x: np.maximum(x[f"last_election_results_{estimand}"] + x.lb, aggregate_nonreporting_votes[f"results_{estimand}"]),
```
lennybronner, line 227: can you add a comment above this explaining why this is necessary. something around the lines of, the gaussian model re-uses the unadjusted bounds to estimate the gaussian for the aggregates. so we need to do this maximum again.

daragold: Added

src/elexmodel/models/BaseElectionModel.py
daragold: Added!

lennybronner (approved):
This looks good to me. But lets hold off merging until after we have added everything to the repo.

PR #4, opened 2022-09-19
Set up repository

emilyliu7321:
**Description:**

- Adds .gitignore, codeowner files
- Adds pre commit, unit test, and integration test workflows
- Bump `elex-solver` version to <2 
- Adds PR template 
- Updated Github labels 

**Test steps:**
1. Unit test and integration test workflows pass


.github/workflows/integration.yml
```diff
@@ -0,0 +1,29 @@
+name: Integration Test
+on: [push, pull_request]
+jobs:
+  test:
+    name: Run integration test
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    strategy:
+      matrix:
+        python-version: [3.8]
```
chriszs, line 10: Should this happen on 3.7, since that's what data-importer is set to?

chriszs (approved):
Looks good. One question.

.github/workflows/integration.yml
emilyliu7321: I think the data importer version should get bumped up instead!

chriszs: Would it make any sense to matrix both versions for now, e.g. 
```suggestion
        python-version: [3.8, 3.7]
```

chriszs: (This way we cover ourselves from introducing breakage into importer while preparing the dependencies for the glorious future.)

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
+name: test
+on: [push, pull_request]
+jobs:
+  test:
+    name: Run unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    strategy:
+      matrix:
+        python-version: [3.8]
```
emilyliu7321, line 10:
```suggestion
        python-version: [3.8, 3.7]
```

chriszs (approved):

.github/workflows/integration.yml
```diff
@@ -0,0 +1,29 @@
+name: Integration Test
+on: [push, pull_request]
```
chriszs, line 2: See other comment.

chriszs:
This is great, thank you! But seems like it's running some tests twice:

Screen Shot 2022-09-19 at 5 28 15 PM

See comment for idea about that.

chriszs (approved):

PR #10, opened 2022-09-26
Add Github Action to automate releases

emilyliu7321:
## Description

- Added Github Action that automatically deploys the package to PyPi and Test PyPi on pushes to `main`

## Jira Ticket
n/a

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,17 @@
+name: Release Stable Version
+on:
+  push:
+    branches:
+      - main
+jobs:
+  - name: Publish package to TestPyPI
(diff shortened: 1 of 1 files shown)


lennybronner (commented):
Is there any good way for me to test this?

emilyliu7321:
@lennybronner I think the best way is to just wait for a release because I also don't want to release any code in develop before it's ready

lennybronner:
> @lennybronner I think the best way is to just wait for a release because I also don't want to release any code in develop before it's ready

ok, will approve for now

lennybronner (approved):

PR #11, opened 2022-09-27
Elex-1303-median error message

daragold:
## Description
We are now writing a warning when the median problem comes up, and using the conformity value that has >.5 weight (first in list) as the median
## Jira Ticket
https://arcpublishing.atlassian.net/jira/software/c/projects/ELEX/boards/1026?modal=detail&selectedIssue=ELEX-1303
## Test Steps
Replicate error and check message appears only in that instance. confirm logger syntax is correct

### DIFF
--- a/src/elexmodel/utils/math_utils.py
+++ b/src/elexmodel/utils/math_utils.py
@@ -1,6 +1,11 @@
 import numpy as np
 from scipy.stats import bootstrap
 
+class ModelClientException(Exception):
+    pass
+
+class ModelNotEnoughSubunitsException(ModelClientException):
+    pass
 
 def compute_inflate(x):
     """
@@ -30,7 +35,12 @@ def weighted_median(x, weights):
 
     # find index of largest x_i where weights are less than or equal 0.5
     weights_cumulative = np.cumsum(weights_sorted)
+    if weights_cumulative[0] > .5:
+        raise ModelNotEnoughSubunitsException(
+            "Lowest weight conformity unit is >50%. Need more data for conformity median."
+        )
     median_index = np.where(weights_cumulative <= 0.5)[0][-1]
+    
 
     # if there is one element where weights are exactly 0.5, median is average
     # otherwise weighted median is the next largest element
--- a/src/elexmodel/utils/math_utils.py
+++ b/src/elexmodel/utils/math_utils.py
@@ -38,10 +38,10 @@ def weighted_median(x, weights):
 
     # find index of largest x_i where weights are less than or equal 0.5
     weights_cumulative = np.cumsum(weights_sorted)
-    #conformity scores are lined up in size order, but each is assigned a 
-    #weight based on unit population. The list is split in half according to
-    #cumulative weights. But if the first element in the list is already over
-    #50% of total weight, there will be nothing in one side of the list.
+    # conformity scores are lined up in size order, but each is assigned a
+    # weight based on unit population. The list is split in half according to
+    # cumulative weights. But if the first element in the list is already over
+    # 50% of total weight, there will be nothing in one side of the list.
     if weights_cumulative[0] > 0.5:
         raise ModelNotEnoughSubunitsException(
             "Lowest weight conformity unit is >50%. Need more data for conformity median."
--- a/src/elexmodel/utils/math_utils.py
+++ b/src/elexmodel/utils/math_utils.py
@@ -1,9 +1,9 @@
 import logging
-
+import warnings
 import numpy as np
 from scipy.stats import bootstrap
 
-LOG = logging.getLogger()
+#LOG = logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


src/elexmodel/utils/math_utils.py
```diff
@@ -2,6 +2,14 @@
 from scipy.stats import bootstrap
 
 
+class ModelClientException(Exception):
```
lennybronner, line 5: we already define a `ModelClientException` in `client.py`. I'm not 100% sure what the implications of doing this twice is. It might be good to move the exceptions to an `exceptions.py` file and then use them in all the places we need them. Curious what you think @emilyliu7321?

lennybronner (commented):
Do you mind putting a short explainer in the description about what caused the error and also adding a unit test for this?

src/elexmodel/utils/math_utils.py
emilyliu7321: I agree with you!

tests/utils/math_utils.py
```diff
@@ -76,6 +80,16 @@ def test_weighted_median():
     assert median == 4
 
 
+def test_weighted_median_error():
+    x = np.array([0, 10, 20])
+    w = np.array([60, 10, 30])
+    w = w / np.sum(w)
+    weights_cumulative = np.cumsum(w)
+    if weights_cumulative[0] >= 0.5:
+        LOG.warning("Warning: smallest conformity value is greater than or equal to half the total weight")
```
lennybronner, line 89: why are we logging in the test?

tests/utils/math_utils.py
```diff
@@ -76,6 +80,16 @@ def test_weighted_median():
     assert median == 4
 
 
+def test_weighted_median_error():
```
lennybronner, line 83: can you add a test with one unit only? that should now work, right? Also please add a test at exactly 50% of the weight.

src/elexmodel/utils/math_utils.py
```diff
@@ -30,7 +34,16 @@ def weighted_median(x, weights):
 
     # find index of largest x_i where weights are less than or equal 0.5
     weights_cumulative = np.cumsum(weights_sorted)
-    median_index = np.where(weights_cumulative <= 0.5)[0][-1]
+    # conformity scores are lined up in size order, but each is assigned a
```
lennybronner, line 37: this function is more general, so could you change the comment to just be about weights?

src/elexmodel/utils/math_utils.py
```diff
@@ -30,7 +34,16 @@ def weighted_median(x, weights):
...
+    # weight based on unit population. The list is split in half according to
+    # cumulative weights. But if the first element in the list is already over
+    # 50% of total weight, there will be nothing in one side of the list.In
+    # that case return the first element
+    if weights_cumulative[0] >= 0.5:
+        LOG.warning("Warning: smallest conformity value is greater than or equal to half the weight")
```
lennybronner, line 43: same with the log message

src/elexmodel/utils/math_utils.py
```diff
@@ -30,7 +34,16 @@ def weighted_median(x, weights):
 
     # find index of largest x_i where weights are less than or equal 0.5
     weights_cumulative = np.cumsum(weights_sorted)
-    median_index = np.where(weights_cumulative <= 0.5)[0][-1]
+    # conformity scores are lined up in size order, but each is assigned a
+    # weight based on unit population. The list is split in half according to
+    # cumulative weights. But if the first element in the list is already over
+    # 50% of total weight, there will be nothing in one side of the list.In
+    # that case return the first element
+    if weights_cumulative[0] >= 0.5:
```
lennybronner, line 42: should this not be greater than?

src/elexmodel/utils/math_utils.py
```diff
@@ -30,7 +34,16 @@ def weighted_median(x, weights):
...
+        LOG.warning("Warning: smallest conformity value is greater than or equal to half the weight")
+        return x_sorted[0]
```
lennybronner, line 44: is there a way to continue to use `median_index`? Like set `median_index` to zero? No, right? since in the most common case we would now return the second element from the left.

src/elexmodel/utils/math_utils.py
daragold: Changed.

src/elexmodel/utils/math_utils.py
daragold: changed to general 'x-value'

emilyliu7321 (approved):
Ran `elexmodel 1996-11-05_USA_G --office_id=P --estimands=turnout --geographic_unit_type=county --pi_method=gaussian --percent_reporting=90` to test the warning statement 👍

PR #13, opened 2022-09-30
1302 save results even with not enough subunits

lennybronner:
## Description
We want to write the results to s3 even if the model run doesn't happen because we don't have enough subunits. This helps with pure results analysis on election night and debugging.

## Jira Ticket
[elex-1302](https://arcpublishing.atlassian.net/browse/ELEX-1302)

## Test Steps
Change `APP_ENV` to `dev` locally. Then this line:
```
elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=county --pi_method=gaussian --percent_reporting 1 --save_output=results
```
should not do anything in `develop` branch, but should update the results file in `dev` bucket in this branch.


emilyliu7321 (approved):
LGTM!

PR #19, opened 2022-10-08
Elex 1509 ecos solver fix when relative weights small

lennybronner:
## Description
Usually we normalize the weights before passing them into the quantile regression solver because that speeds the solution up. However, it seems that if the normalized smallest weight is too close to zero, then that throws a warning about inaccurate solutions or occasionally throws an error. 

Now, if a warning or an error is thrown we catch it and re-run the quantile regression solver with `normalize_weights=False`, this is slower but should be more accurate.

## Jira Ticket
[elex-1509](https://arcpublishing.atlassian.net/browse/ELEX-1509)

## Test Steps
This would throw an uncaught warning or sometimes error out. Now no longer does that:
```
elexmodel 2022-11-08_USA_G --office_id=S --aggregates=postal_code --pi_method=nonparametric --geographic_unit_type=county --percent_reporting_threshold=99 --prediction_intervals=0.9 --historical --percent_reporting=20 --estimands=dem
```

### DIFF
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -1,12 +1,16 @@
 import math
 from collections import namedtuple
+import warnings
+warnings.filterwarnings('error', category=UserWarning, module='cvxpy')
+import logging
 
+import cvxpy
 import numpy as np
 from elexsolver.QuantileRegressionSolver import QuantileRegressionSolver
 
 PredictionIntervals = namedtuple("PredictionIntervals", ["lower", "upper", "conformalization"], defaults=(None,) * 3)
 
-
+LOG = logging.getLogger(__name__)
 class BaseElectionModel(object):
     def __init__(self, model_settings={}):
         self.qr = QuantileRegressionSolver(solver="ECOS")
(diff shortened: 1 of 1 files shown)


src/elexmodel/models/BaseElectionModel.py
```diff
@@ -33,7 +40,16 @@ def fit_model(self, model, df_X, df_y, tau, weights, normalize_weights):
         X = df_X.loc[:, columns_to_keep].values
         y = df_y.values
         weights = weights.values
-        model.fit(X, y, tau_value=tau, weights=weights, normalize_weights=normalize_weights)
+
+        # normalizing weights speed up solution by a lot. However, if the relative size
+        # of the smallest weight is too close to zero, it can lead to numerical instability
+        # where the solver either throws a warning for inaccurate solution or breaks entirely
+        # in that case, we catch the error and warning and re-run with normalize_weights false
+        try:
```
anthonyjpesce, line 48: Should your example command fail/error from the develop branch? Seems to run OK for me

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -33,7 +40,16 @@ def fit_model(self, model, df_X, df_y, tau, weights, normalize_weights):
...
+            model.fit(X, y, tau_value=tau, weights=weights, normalize_weights=normalize_weights)
+        except (UserWarning, cvxpy.error.SolverError):
+            LOG.warning("Warning: solution was inaccurate or solver broke. Re-running with normalize_weights=False.")
```
anthonyjpesce, line 51: I'm not getting this log when I run your example command

anthonyjpesce: Never mind, this was user error.

anthonyjpesce (approved):

PR #28, opened 2022-12-12
Fix/write district in bounds

lennybronner:
## Description
On election night in November we realized that the district column of the bounds produced by the GaussianModel of the `precinct-district` model with gaussian prediction interval aggregation method were missing. 

After investigating, I realized that the column should have been missing because the rows in the `bounds.csv` file represent the gaussian model for all precincts (nation-wide) and the second row for all precincts in Virginia (both of which are not district specific). The reason that the `postal_code` for both was `VA` was because we were overwriting that column with data from the conformalization set.

I've removed the code that does the overwriting. But I am curious what that code was supposed to do initially? 

I also added default type for `contentType` as `text/csv` when writing CSVs to s3 in the `S3CsvUtil`. The default set for anything written to s3 in the `S3Util` was as `contentType` `application/json`. And so downloading all these CSVs from the s3 turned the files into json.

## Jira Ticket
[elex-1980](https://arcpublishing.atlassian.net/browse/ELEX-1980)

## Test Steps
In develop branch change line 94 in `cli.py` to `data_handler.shuffle(upweight={'postal_code': {'VA': 100000}})` and run:
```
elexmodel 2022-11-08_USA_G --office_id=H_precinct-district --estimands=gop --geographic_unit_type=precinct-district --pi_method=gaussian --historical --percent_reporting=1 --aggregates=district --save_output=conformalization
```
In `bounds.csv` (you will need to check in `2020-11-03` since historical is `True`) all `postal_code` are overwritten to be `VA`. If you rerun this in the branch the first gaussian bound should have empty postal_code (it is for all precincts, regardless of state).

### DIFF
--- a/src/elexmodel/distributions/GaussianModel.py
+++ b/src/elexmodel/distributions/GaussianModel.py
@@ -3,7 +3,6 @@ import pandas as pd
 
 from elexmodel.handlers import s3
 from elexmodel.utils import math_utils, pandas_utils
-from elexmodel.utils.constants import AGGREGATE_ORDER
 from elexmodel.utils.file_utils import S3_FILE_PATH, TARGET_BUCKET, convert_df_to_csv
 
 
--- a/src/elexmodel/distributions/GaussianModel.py
+++ b/src/elexmodel/distributions/GaussianModel.py
@@ -217,10 +217,6 @@ class GaussianModel:
                     aggregate,
                     alpha,
                 )
-                # Write columns for preceding aggregates
-                for agg in AGGREGATE_ORDER[: AGGREGATE_ORDER.index(aggregate[-1])]:
-                    if agg in conformalization_data:
-                        gaussian_bounds[agg] = conformalization_data[agg]
                 # Write bounds
                 self._write_gaussian_bounds(
                     gaussian_bounds,
--- a/src/elexmodel/handlers/s3.py
+++ b/src/elexmodel/handlers/s3.py
@@ -66,6 +66,7 @@ class S3CsvUtil(S3Util):
         """
         if not filename.endswith(".csv"):
             filename = f"{filename}.csv"
+        kwargs.setdefault("ContentType", "text/csv")
         super().put(filename, data, **kwargs)
 
     def get(self, filename, load=True, **kwargs):


lennybronner:
I am not sure what this code was supposed to do:
```
# Write columns for preceding aggregates
for agg in AGGREGATE_ORDER[: AGGREGATE_ORDER.index(aggregate[-1])]:
  if agg in conformalization_data:
    gaussian_bounds[agg] = conformalization_data[agg]
```

emilyliu7321:
> I am not sure what this code was supposed to do:
> 
> ```
> # Write columns for preceding aggregates
> for agg in AGGREGATE_ORDER[: AGGREGATE_ORDER.index(aggregate[-1])]:
>   if agg in conformalization_data:
>     gaussian_bounds[agg] = conformalization_data[agg]
> ```

@lennybronner This change was introduced as part of the PR to write conformalization data and gaussian bounds to S3, and we wanted to write the gaussian bounds at the top level of recursion or else the aggregate group information wouldn't get included in the uploaded csvs. (the aggregate group here would typically be `postal_code` but varies based on what parameters we're running the model on)

lennybronner:
> > I am not sure what this code was supposed to do:
> > ```
> > # Write columns for preceding aggregates
> > for agg in AGGREGATE_ORDER[: AGGREGATE_ORDER.index(aggregate[-1])]:
> >   if agg in conformalization_data:
> >     gaussian_bounds[agg] = conformalization_data[agg]
> > ```
> 
> @lennybronner This change was introduced as part of the PR to write conformalization data and gaussian bounds to S3, and we wanted to write the gaussian bounds at the top level of recursion or else the aggregate group information wouldn't get included in the uploaded csvs. (the aggregate group here would typically be `postal_code` but varies based on what parameters we're running the model on)

I guess the issue is that we are overwriting the column in the district case.`gaussian_bounds` has one row per gaussian model generated. So for example looks like this:
```
    var_inflate  mu_lower_bound  mu_upper_bound  sigma_lower_bound  sigma_upper_bound postal_code district
0      0.004250       -0.158309       -0.264542           1.703662           1.699356         NaN      NaN
1      0.143908       -0.051613       -0.371239           0.617041           0.619079          AL      NaN
2      0.185278       -0.221239       -0.201613           0.803454           0.805127          AR      NaN
3      0.043693       -0.418895       -0.003957           4.863760           4.855206          CA      NaN
4      0.184100       -0.158309       -0.264542           0.337153           0.336895          FL      NaN
5      0.040756       -0.058091       -0.364760           0.881531           0.881208          IL      NaN
6      0.072089       -0.117647       -0.305205           0.139839           0.139629          IN      NaN
7      0.063192       -0.177835       -0.245017           1.261995           1.260382          MI      NaN
8      0.087155       -0.132701       -0.290150           0.137031           0.137073          MN      NaN
9      0.050644       -0.138182       -0.284670           1.856130           1.844893          NY      NaN
10     0.054676       -0.105455       -0.317397           0.909546           0.908326          OH      NaN
11     0.051990       -0.099788       -0.323064           0.953529           0.948731          PA      NaN
12     0.063961       -0.230609       -0.192242           3.707269           3.709996          TX      NaN
13     0.058118       -0.325758       -0.097094           0.478134           0.474861          WA      NaN
14     0.157769       -0.128415       -0.294436           0.090361           0.091501          WI      NaN
```
(all districts are `NA` cuz these are bounds for the states only). 

The conformalization data though has one row per unit in the conformalization set, so it usually has way more rows than the gaussian bounds. So we are overwriting the columns in the gaussian model dataframe with the first few rows from the conformalization data.

emilyliu7321:
@lennybronner I think you're right, I tested this with the old code that adds a column with this command:

`elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=county --pi_method=gaussian --aggregates="postal_code" --aggregates=county_classification --save_output=conformalization`

and whether or not that code snippet of adding columns was included, the gaussian bound data seems to be the same anyways

emilyliu7321 (approved):

PR #38, opened 2023-04-05
Elex 2329 move fixed effect to model

lennybronner:
## Description
We are moving the fixed effect generation from the `CombinedDataHandler` to a `Featurizer` which is invoked in the `BaseElectionModel`. This lets us generate fixed effect for each model fitting independently (ie. for the median, and for the prediction intervals), without having to worry about whether fixed effects may not appear in the training and conformalization set. It also tracks more closely with how model fitting is done in other packages, where categorical variables are handled for every model run independently. 

This also let's us remove some of the code we wrote to handle weird edge cases with the fixed effects. We now no longer drop zero or duplicate columns before fitting the model since that happening is now purely user error and should cause the model to error.

This PR also fixes a bug with centering the covariates. In the past, we were centering based on the column center of each dataframe, now we are computing the column center for all subunits when computing the mean.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2329

## Test Steps
Unit tests have been added so `tox`

Also run the 2020 election using this model branch and the model will no longer error out at 22:23pm.


src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -0,0 +1,123 @@
...
+        self.add_intercept = add_intercept
+
+        if self.center_features:
+            self._center_features(new_fitting_data)
+
+        if self.add_intercept:
+            self._add_intercept(new_fitting_data)
+
+        if len(self.fixed_effects) > 0:
+            # drop_first is True for fitting_data (reporting_units) since we want to avoid the design matrix with
```
daragold, line 74: Can you clarify here in this comment all the places we have "fitting data" (I think its just reporting data and the training data df in the conform step?). Also we should write somewhere which fitting data dfs and prediction dfs have the same features, have a subset of each others features etc.

src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -0,0 +1,123 @@
...
+            self._center_features(new_heldout_data)
+
+        if self.add_intercept:
+            self._add_intercept(new_heldout_data)
+
+        if len(self.fixed_effects) > 0:
+            missing_expanded_fixed_effects = []
+            new_heldout_data = self._expand_fixed_effects(new_heldout_data, drop_first=False)
+            # if all units from one fixed effect are reporting they will not appear in the heldout_data (nonreporting_units) and won't
+            # get a column when we expand the fixed effects on that dataframe. Therefore we add those columns with zero
```
daragold, line 106: similar here - can you clarfy that non-reporting units is one example of heldout_data? I think the only other is the conformalization data in the CI training/fitting?

src/elexmodel/handlers/data/Featurizer.py
lennybronner: ok, just added that to the functions docstrings, let me know what you think.
Also, write out which dataframes have the same features (and subsets etc.) when invoking the featurizer in `BaseElectionModel`

src/elexmodel/handlers/data/Featurizer.py
lennybronner: Same as above, let me know

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -234,7 +226,9 @@ def get_unit_prediction_interval_bounds(self, reporting_units, nonreporting_unit
         conformalization_data["lower_bounds"] = conformalization_lower_bounds
 
         # apply lower/upper models to nonreporting data
-        nonreporting_units_features = nonreporting_units[self.features].values
+        # now the features of teh nonreporting_units will be the same as the train_data features
+        # they might differ slightly from the features used when fitting the median prediction
+        nonreporting_units_features = self.featurizer.featurize_heldout_data(nonreporting_units)
```
daragold, line 231: Just want to clarify: this step now means the nonreporting units (which previously had their fixed effects matched to the reporting units) is now having them re-matched to the training/comform fixed effects. So it is still the case that for a non-reporting unit, the fixed effects used to get its point prediction can be different from those used to get its 5th/95th percentile points, correct? I dont think thats necessarily bad, just want to confirm this is what you intended.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -205,7 +193,9 @@ def get_unit_prediction_interval_bounds(self, reporting_units, nonreporting_unit
 
         # specifying self.features extracts the correct columns and makes sure they are in the correct
         # order. Necessary when fitting and predicting on the model.
-        train_data_features = train_data[self.features]
+        # the fixed effects in train_data will be a subset of the fixed effect of reporting_units since all
+        # units from one fixed effect category might be in the conformalization data.
+        train_data_features = self.featurizer.featurize_fitting_data(train_data)
```
daragold, line 198: I think there's an issue with your center_features flag. Right now you have it that those means (which are computed once when the Featurizer is initiated) get subtracted from the features in every call of the your featurizing functions. (I don't see you setting the flag to false in any of the calls). Don't we just want to do this once? Like maybe when you compute the means for centuring in line 52, you should also subtract the means from all the units (reporting+nonreporting) then never do it again? Is that right?

daragold:
> ## Description
> We are moving the fixed effect generation from the `CombinedDataHandler` to a `Featurizer` which is invoked in the `BaseElectionModel`. This lets us generate fixed effect for each model fitting independently (ie. for the median, and for the prediction intervals), without having to worry about whether fixed effects may not appear in the training and conformalization set. It also tracks more closely with how model fitting is done in other packages, where categorical variables are handled for every model run independently.
> 
> This also let's us remove some of the code we wrote to handle weird edge cases with the fixed effects. We now no longer drop zero or duplicate columns before fitting the model since that happening is now purely user error and should cause the model to error.
> 
> This PR also fixes a bug with centering the covariates. In the past, we were centering based on the column center of each dataframe, now we are computing the column center for all subunits when computing the mean.
> 
> ## Jira Ticket
> https://arcpublishing.atlassian.net/browse/ELEX-2329
> 
> ## Test Steps
> Unit tests have been added so `tox`
> 
> Also run the 2020 election using this model branch and the model will no longer error out at 22:23pm.

This looks so great! Really good reorg. Some comments below.
Also can you please add some error calls/ assert statements making sure the order of the fixed effects columns are the same in all the places they need to be (i.e. non-reporting when its featurized to be training data etc.etc.). Maybe the check is to make sure the column  vectors between  all fitted and prediction dataframes are identical post-featurizing or something? not sure, but I think we def need this.

src/elexmodel/models/BaseElectionModel.py
lennybronner: Yes, but this was previously already the case. We never made sure that the train data had the same columns as the reporting data, so it was always possible that the model was slightly different.

src/elexmodel/models/BaseElectionModel.py
lennybronner: hmm, I'm not sure that is true. We pass in the dataframe that we want to featurize, copy it and then subtract the means. So we're never subtracting the means from `reporting_units` (we're subtracting it from a copy of reporting_units). So when we pass in `train_data` (the subset of `reporting_units`) those columns have never been touched.

lennybronner:
> > ## Description
> > We are moving the fixed effect generation from the `CombinedDataHandler` to a `Featurizer` which is invoked in the `BaseElectionModel`. This lets us generate fixed effect for each model fitting independently (ie. for the median, and for the prediction intervals), without having to worry about whether fixed effects may not appear in the training and conformalization set. It also tracks more closely with how model fitting is done in other packages, where categorical variables are handled for every model run independently.
> > This also let's us remove some of the code we wrote to handle weird edge cases with the fixed effects. We now no longer drop zero or duplicate columns before fitting the model since that happening is now purely user error and should cause the model to error.
> > This PR also fixes a bug with centering the covariates. In the past, we were centering based on the column center of each dataframe, now we are computing the column center for all subunits when computing the mean.
> > ## Jira Ticket
> > https://arcpublishing.atlassian.net/browse/ELEX-2329
> > ## Test Steps
> > Unit tests have been added so `tox`
> > Also run the 2020 election using this model branch and the model will no longer error out at 22:23pm.
> 
> This looks so great! Really good reorg. Some comments below. Also can you please add some error calls/ assert statements making sure the order of the fixed effects columns are the same in all the places they need to be (i.e. non-reporting when its featurized to be training data etc.etc.). Maybe the check is to make sure the column vectors between all fitted and prediction dataframes are identical post-featurizing or something? not sure, but I think we def need this.

I'm not opposed to doing this (though I would want to think of the best way), but I am curious how you think that might happen?

lennybronner:
ok, I added a unit test to test whether that is true. Let me know what you think.

lennybronner:
> > ## Description
> > We are moving the fixed effect generation from the `CombinedDataHandler` to a `Featurizer` which is invoked in the `BaseElectionModel`. This lets us generate fixed effect for each model fitting independently (ie. for the median, and for the prediction intervals), without having to worry about whether fixed effects may not appear in the training and conformalization set. It also tracks more closely with how model fitting is done in other packages, where categorical variables are handled for every model run independently.
> > This also let's us remove some of the code we wrote to handle weird edge cases with the fixed effects. We now no longer drop zero or duplicate columns before fitting the model since that happening is now purely user error and should cause the model to error.
> > This PR also fixes a bug with centering the covariates. In the past, we were centering based on the column center of each dataframe, now we are computing the column center for all subunits when computing the mean.
> > ## Jira Ticket
> > https://arcpublishing.atlassian.net/browse/ELEX-2329
> > ## Test Steps
> > Unit tests have been added so `tox`
> > Also run the 2020 election using this model branch and the model will no longer error out at 22:23pm.
> 
> This looks so great! Really good reorg. Some comments below. Also can you please add some error calls/ assert statements making sure the order of the fixed effects columns are the same in all the places they need to be (i.e. non-reporting when its featurized to be training data etc.etc.). Maybe the check is to make sure the column vectors between all fitted and prediction dataframes are identical post-featurizing or something? not sure, but I think we def need this.

ok, I added a unit test to test whether that is true. Let me know what you think.

src/elexmodel/handlers/data/Featurizer.py
daragold: Cool -- I think we can then drop the reference to  "(reporting units)" in line 74 or just say e.g. reporting units. Sorry to be anal.

src/elexmodel/handlers/data/Featurizer.py
daragold: same

src/elexmodel/models/BaseElectionModel.py
daragold: cool

src/elexmodel/models/BaseElectionModel.py
daragold: ah ok -- yea i think this is fine then

src/elexmodel/handlers/data/Featurizer.py
lennybronner: no worries and done!

src/elexmodel/handlers/data/Featurizer.py
lennybronner: done!

daragold (approved):

PR #39, opened 2023-04-17
elex-2333-remove-nonreporting-residuals-column

daragold:
## Description
There was a blank residuals column being added to the nonreporting_units matrix that was unused. This caused confusion when debugging issues with reporting and non-reporting units. Nonreporting residuals columns do not get fed into the solver (obviously, because they don't exist and its a non-reporting unit).

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2333

## Test Steps
Run model under several settings to make sure nothing breaks. Gaussian v. nonparametric and feature v no feature. And in a multi-state run, with and without fixed effects.

### DIFF
--- a/src/elexmodel/handlers/data/CombinedData.py
+++ b/src/elexmodel/handlers/data/CombinedData.py
@@ -1,5 +1,3 @@
-import numpy as np
-
 from elexmodel.handlers import s3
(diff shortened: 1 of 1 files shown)


lennybronner:
Do you mind adding in a link to the jira ticket and also how to test that this change doesn't break anything?

lennybronner (approved):

PR #40, opened 2023-04-17
elex-2334-remove-total-voters-columns

daragold:
## Description
"total_voters_{estimand}" has been removed everywhere. Instead we add in the additional voter (needed to avoid division by 0) into the "last_election_results_{estimand}" columns directly.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2334

## Test Steps
Run model and make sure nothing breaks. Try both gaussian and nonparametric, and no fixed effects v fixed effects = 'postal_code' (this is in a multi-state election test)

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -350,8 +350,7 @@ def test_get_estimates_fully_reporting(model_client, va_governor_county_data, va
     data = data_handler.get_percent_fully_reported(100)
 
     preprocessed_data = va_governor_county_data.copy()
-    preprocessed_data["last_election_results_turnout"] = preprocessed_data["baseline_turnout"].copy()
-    preprocessed_data["total_voters_turnout"] = preprocessed_data["last_election_results_turnout"] + 1
+    preprocessed_data["last_election_results_turnout"] = preprocessed_data["baseline_turnout"].copy() + 1
 
     result = model_client.get_estimates(
         data,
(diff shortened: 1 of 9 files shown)


tests/test_client.py
```diff
@@ -585,8 +579,8 @@ def test_get_estimates_no_subunits_reporting(model_client, va_governor_county_da
     data = data_handler.get_percent_fully_reported(0)
 
     preprocessed_data = va_governor_county_data.copy()
-    preprocessed_data["last_election_results_turnout"] = preprocessed_data["baseline_turnout"].copy()
-    preprocessed_data["total_voters_turnout"] = preprocessed_data["last_election_results_turnout"] + 1
+    preprocessed_data["last_election_results_turnout"] = preprocessed_data["baseline_turnout"].copy() + 1
+    #   preprocessed_data["last_election_results_turnout"] = preprocessed_data["last_election_results_turnout"] + 1
```
lennybronner, line 583: do you mind just removing this line to avoid confusion

lennybronner (commented):
This looks great! Similar to elex-2333, please add the jira ticket, add testing instructions and update the PR name.

tests/test_client.py
daragold: done

lennybronner (approved):

PR #41, opened 2023-04-18
elex-2331-nan-warnings

daragold:
## Description
We want to throw a warning when nan values are fed into the solver. This now checks reporting and nonreporting data that is then fitted. Note, we don't need to check before the additional solver fittings during the conformalization steps, as this is just using subsets of the same data.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2331

## Test Steps
New units tests - run tox.
(also sanity check placement of new warnings)

### DIFF
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -63,13 +63,13 @@ class BaseElectionModel(object):
         # of the reporting units, which are all being tested here.
 
         if reporting_units_features.isnull().values.any():
-            warnings.warn("NaN values in reporting_units_features df")
+            LOG.warning("Warning: NaN values in reporting_units_features")
 
         if reporting_units_residuals.isnull().values.any():
-            warnings.warn("NaN values in reporting_units_residuals")
-
+            LOG.warning("Warning: NaN values in reporting_units_residuals")
+        print(nonreporting_units_features.head())
         if nonreporting_units_features.isnull().values.any():
-            warnings.warn("NaN values in nonreporting_units_features")
+            LOG.warning("Warning: NaN values in nonreporting_units_features")
 
         self.fit_model(self.qr, reporting_units_features, reporting_units_residuals, 0.5, weights, True)
 
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -58,6 +58,19 @@ class BaseElectionModel(object):
         weights = reporting_units[f"last_election_results_{estimand}"]
         reporting_units_residuals = reporting_units[f"residuals_{estimand}"]
 
+        # check if any Nan values are getting fed into the solver. We don't need to repeat this
+        # for the additional lower/upper bound quantile fittings, as they're just using subsets
+        # of the reporting units, which are all being tested here.
+
+        if reporting_units_features.isnull().values.any():
+            warnings.warn("NaN values in reporting_units_features df")
+
+        if reporting_units_residuals.isnull().values.any():
+            warnings.warn("NaN values in reporting_units_residuals")
+
+        if nonreporting_units_features.isnull().values.any():
+            warnings.warn("NaN values in nonreporting_units_features")
+
         self.fit_model(self.qr, reporting_units_features, reporting_units_residuals, 0.5, weights, True)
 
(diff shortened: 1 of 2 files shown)


tests/models/test_base_election_model.py
```diff
@@ -25,6 +26,42 @@ def test_fit_model():
...
+    df["demo_feature"] = np.random.choice([100, 200], size=len(df))
+    df1 = df[:1000].copy()
+    df1["reporting"] = 1
+    df1[f"residuals_{estimand}"] = np.random.choice([0, 1], size=len(df1))
+    df2 = df[1000:2000]
+    df2["reporting"] = 0
+
+    model.get_unit_predictions(df1, df2, estimand)
+    with pytest.warns(None):
+        model.get_unit_predictions(df1, df2, estimand)
```
lennybronner, line 53: This is cool. I didn't know about `pytest.warns`. Just to make sure I understand, this is testing that no warning is raised?

lennybronner (commented):
I think it makes more sense to move the checks to `fit_model`, since then it would happen for any group of covariates and response variables that we pass to the model, what do you think? At the moment we aren't running this check for confidence intervals for example.

tests/models/test_base_election_model.py
daragold: Correct

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -58,6 +58,19 @@ def get_unit_predictions(self, reporting_units, nonreporting_units, estimand):
         weights = reporting_units[f"last_election_results_{estimand}"]
         reporting_units_residuals = reporting_units[f"residuals_{estimand}"]
 
+        # check if any Nan values are getting fed into the solver. We don't need to repeat this
+        # for the additional lower/upper bound quantile fittings, as they're just using subsets
+        # of the reporting units, which are all being tested here.
+
+        if reporting_units_features.isnull().values.any():
+            warnings.warn("NaN values in reporting_units_features df")
```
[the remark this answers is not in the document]
daragold: Done

lennybronner:
I'm getting these warnings when I run the new unit tests:
```
tests/models/test_base_election_model.py::test_nan_data_warning
  /Users/bronnerl/Projects/elex-live-model/tests/models/test_base_election_model.py:43: SettingWithCopyWarning:
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead

  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df[f"last_election_results_{estimand}"] = df[f"results_{estimand}"] - 10

tests/models/test_base_election_model.py::test_nan_data_warning
  /Users/bronnerl/Projects/elex-live-model/tests/models/test_base_election_model.py:44: SettingWithCopyWarning:
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead

  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df["demo_feature"] = np.random.choice([100, 200], size=len(df))

tests/models/test_base_election_model.py::test_nan_data_warning
  /Users/bronnerl/Projects/elex-live-model/tests/models/test_base_election_model.py:49: SettingWithCopyWarning:
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead

  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df2["reporting"] = 0

tests/models/test_base_election_model.py::test_nan_data_warning
  /Users/bronnerl/Projects/elex-live-model/.tox/py3/lib/python3.10/site-packages/_pytest/python.py:194: PytestRemovedIn8Warning: Passing None has been deprecated.
  See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests for alternatives in common use cases.
    result = testfunction(**testargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

PR #43, opened 2023-04-21
feat: add selection of fixed effects

jchaskell:
## Description
Allows you to select which fixed effects are included in the model. All other values for that variable are coded as "other". You do so with a fixed_effects parameter that looks like:
```
'{"state_postal": ["PA", "CA"], "county_classification": ["all"]}'
```
(it needs to have the quotes in that order on the CLI - just pass as a dictionary to the client)

This change is _kinda_ backward compatible in that you can still pass `Featurizer` a list of fixed effects and it will assume that you want all FEs for everything in the list. Additionally, you can pass 1 set of fixed effects in the old way to the cli, *but* you cannot pass multiple `--fixed_effects` parameters.

## Jira Ticket
elex-2375

## Test Steps
```
elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=precinct --aggregates=county_classification --aggregates=postal_code --fixed_effects='{"county_classification": ["nova", "southside"]}' --percent_reporting 10 --features=ethnicity_european --features=ethnicity_hispanic_and_portuguese

elexmodel 2021-01-05_GA_G --office_id=S_precinct --estimands=dem --geographic_unit_type=precinct --pi_method=gaussian --percent_reporting=20 --fixed_effects='{"county_fips": ["13311", "13319"], "county_classification": ["all"]}'

```

### DIFF
--- a/src/elexmodel/handlers/data/Featurizer.py
+++ b/src/elexmodel/handlers/data/Featurizer.py
@@ -48,16 +48,15 @@ class Featurizer(object):
         """
         Convert fixed effect columns into dummy variables.
         """
-        # we concatenate the dummy variables with the original fixed effects, since we need the original fixed
-        # effect columns for aggregation.
         original_fixed_effect_columns = df[self.fixed_effect_cols]
         # set non-included values to other as needed
+        fe_df = df.copy()
         for fe, params in self.fixed_effect_params.items():
             if "all" not in params:
-                df[fe] = np.where(~df[fe].isin(params), "other", df[fe])
+                fe_df[fe] = np.where(~fe_df[fe].isin(params), "other", fe_df[fe])
 
         expanded_fixed_effects = pd.get_dummies(
-            df,
+            fe_df,
             columns=self.fixed_effect_cols,
             prefix=self.fixed_effect_cols,
             prefix_sep="_",
@@ -74,6 +73,8 @@ class Featurizer(object):
                 else:
                     cols_to_drop.append(relevant_cols[0])
 
+        # we concatenate the dummy variables with the original fixed effects, since we need the original fixed
+        # effect columns for aggregation.
         return pd.concat(
             [
                 original_fixed_effect_columns,
--- a/src/elexmodel/handlers/data/Featurizer.py
+++ b/src/elexmodel/handlers/data/Featurizer.py
@@ -9,12 +9,18 @@ class Featurizer(object):
 
     def __init__(self, features, fixed_effects):
         self.features = features
-        if type(fixed_effects) == "list":
+        if isinstance(fixed_effects, list):
             self.fixed_effect_cols = fixed_effects
             self.fixed_effect_params = {fe: ["all"] for fe in fixed_effects}
         else:
-            self.fixed_effect_params = fixed_effects
             self.fixed_effect_cols = list(fixed_effects.keys())
+            self.fixed_effect_params = {}
+            for fe, params in fixed_effects.items():
+                if params == "all":
+                    self.fixed_effect_params[fe] = ["all"]
+                else:
+                    self.fixed_effect_params[fe] = params
+
         self.expanded_fixed_effects = []
         self.complete_features = None
         self.column_means = None
@@ -47,7 +53,7 @@ class Featurizer(object):
         original_fixed_effect_columns = df[self.fixed_effect_cols]
         # set non-included values to other as needed
         for fe, params in self.fixed_effect_params.items():
-            if params != "all" or "all" not in params:
+            if "all" not in params:
                 df[fe] = np.where(~df[fe].isin(params), "other", df[fe])
 
         expanded_fixed_effects = pd.get_dummies(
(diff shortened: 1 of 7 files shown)


src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -33,27 +44,34 @@ def _center_features(self, df):
     def _add_intercept(self, df):
         df["intercept"] = 1
 
-    def _expand_fixed_effects(self, df, drop_first):
+    def _expand_fixed_effects(self, df: pd.DataFrame, drop_first: bool) -> pd.DataFrame:
```
lennybronner, line 47: do you think it's worth adding this to all the functions? or slowly as we update code keep on adding?

src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -9,7 +9,18 @@ class Featurizer(object):
...
         self.features = features
-        self.fixed_effects = fixed_effects
+        if isinstance(fixed_effects, list):
+            self.fixed_effect_cols = fixed_effects
+            self.fixed_effect_params = {fe: ["all"] for fe in fixed_effects}
+        else:
+            self.fixed_effect_cols = list(fixed_effects.keys())
+            self.fixed_effect_params = {}
+            for fe, params in fixed_effects.items():
+                if params == "all":
```
lennybronner, line 19: is this in case someone inputs the parameter as:
```
{'postal_code': 'all'}
```
instead of
```
{'postal_code': ['all']}
```

src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -33,27 +44,34 @@ def _center_features(self, df):
...
+
+        expanded_fixed_effects = pd.get_dummies(
+            fe_df, columns=self.fixed_effect_cols, prefix=self.fixed_effect_cols, prefix_sep="_", dtype=np.int64
+        )
+
+        # drop first column or "other" column if drop_first is true
+        cols_to_drop = []
+        if drop_first:
+            for fixed_effect in self.fixed_effect_cols:
+                relevant_cols = [col for col in expanded_fixed_effects.columns if col.startswith(fixed_effect)]
```
lennybronner, line 66: the original fixed effect column is still in the `expanded_fixed_effects` dataframe, right? And the relevant columns shouldn't include the original fixed effect column, so should it not be `startswith(fixed_effect + '_')`?

lennybronner (commented):
I'm curious about the backward compatibility. Is there any particular reason to do that? Is it poor form not to?

src/elexmodel/handlers/data/Featurizer.py
jchaskell: Ha, yeah the latter is what I was thinking and decided to just do that here. We could do a ticket at some point to add all the types but that would be a lot.

src/elexmodel/handlers/data/Featurizer.py
jchaskell: Yes

src/elexmodel/handlers/data/Featurizer.py
jchaskell: The original column gets dropped - that's why we have the concatenation of the two dataframes at the end (to preserve the original column)

jchaskell: > I'm curious about the backward compatibility. Is there any particular reason to do that? Is it poor form not to?

Hmm I think it depends in terms of poor form. For our purposes, it probably doesn't matter. It was a combination of it being relatively easy to do + me being lazy and not wanting to update a million tests + also thinking that it wouldn't break people's test commands that they have. 

Do you think I should get rid of that? It does make things slightly more complicated but I tried to minimize that.

But this reminds me, I should update the docs at least.

jchaskell: It also means we don't have to update the data importer calls but 🤞 we won't be using the model calls in data importer anymore....

src/elexmodel/handlers/data/Featurizer.py
lennybronner: that sounds good to me!

src/elexmodel/handlers/data/Featurizer.py
lennybronner: > The original column gets dropped - that's why we have the concatenation of the two dataframes at the end (to preserve the original column)

oh right, thanks!

lennybronner: > > I'm curious about the backward compatibility. Is there any particular reason to do that? Is it poor form not to?
> 
> Hmm I think it depends in terms of poor form. For our purposes, it probably doesn't matter. It was a combination of it being relatively easy to do + me being lazy and not wanting to update a million tests + also thinking that it wouldn't break people's test commands that they have.
> 
> Do you think I should get rid of that? It does make things slightly more complicated but I tried to minimize that.
> 
> But this reminds me, I should update the docs at least.

I don't mind it. But I wonder if we should not have the dictionary allow for an `all` case then? And the old way just being the way of specifying all fixed effects? mostly I just don't love having multiple ways of doing the same thing, which I am worried might cause confusion. but I do not feel strongly about this, so happy to go with this now.

jchaskell: That doesn't entirely work because because you can't submit multiple FE via the cli then. I did only specify the new way in the docs.

lennybronner: ok

lennybronner (approved):

PR #42, opened 2023-04-20
Elex 2376 add regularization

lennybronner:
## Description
Added a `lambda_` parameter to the CLI and to the client that allows us to specify regularization for our quantile regression model. This will hopefully let us fix situations in which our model is overfit because of fixed effects.

Also added a variable to explicitly set the intercept. This is done for both the `Featurizer` when we add the intercept and for the solver, to avoid regularizing the intercept coefficient.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2376

## Test Steps
Make sure you have elex-solver v1.1.0 installed:
```
pip install elex-solver --upgrade
```

`tox`

First run a model with no features and no fixed effects:
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=turnout --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=5 --aggregates=postal_code
```

Then run a model with fixed effects but no regularization:
```
 elexmodel 2020-11-03_USA_G --office_id=P --estimands=turnout --geographic_unit_type=county --pi_method=nonparametric  --percent_reporting=5 --aggregates=postal_code --fixed_effects=postal_code
```

Then run a model fixed effects and very high regularization:
```
 elexmodel 2020-11-03_USA_G --office_id=P --estimands=turnout --geographic_unit_type=county --pi_method=nonparametric  --percent_reporting=5 --aggregates=postal_code --fixed_effects=postal_code --lambda=1e4
```

You should notice that the output of the last run (with lambda) will be a lot more similar to the run with no fixed effects than with fixed effects because we are regularizing the fixed effect coefficients to zero.

PR #44, opened 2023-04-27
fix checking new fixed effect inputs

lennybronner:
## Description
Fixed small bug where we were checking whether a list was in another list instead of whether an element of the first list was in the other list. This was always false, and causing us to not be able to run the model with fixed effects if passed in through the normal way.

## Test Steps
`tox`
Added a unit test for the dictionary case. This was not caught because the first unit tests in `test_client` are not being run, the fix for that is in this PR: https://github.com/washingtonpost/elex-live-model/pull/42

### DIFF
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -77,7 +77,7 @@ class ModelClient(object):
             ]
         else:
             invalid_fixed_effects = [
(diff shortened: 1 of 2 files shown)


daragold (approved):

jchaskell (approved):
lgtm - thank you!!

PR #54, opened 2023-06-12
Remove model settings

rishasurana:
## Description
We are passing the model settings dictionary into the constructor. We should just use `self.model_settings` when passing the settings to GaussianModel instead of having to pass them `get_aggregate_prediction_intervals` when the function is called. 

Task: Remove model_settings as a parameter for all implementations of `get_aggregate_prediction_intervals`.

Note: I also made these changes within the Nonparametric election model in order to ensure that combined unit tests run smoothly.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2264

## Test Steps
1. `pip install elex-model`
2. `pip install -r requirements.txt`
3. `pip install -r requirements-dev.txt`
4. `Create a .env file with the variables specified in README`
5. `tox`

### DIFF
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -306,7 +306,6 @@ class ModelClient(object):
                         alpha,
                         alpha_to_unit_prediction_intervals[alpha].conformalization,
                         estimand,
-                        model_settings,
                     )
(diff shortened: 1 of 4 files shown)


jchaskell (approved):
looks good to me!

PR #56, opened 2023-06-15
ELEX-2418-find-optimal-lambda

rishasurana:
## Description
We can pass a regularization constant to our model `lambda_`, but on election night we don’t know what the optimal `lambda_` is. Add a feature to the client that take options for `lambda_`, the features and fixed effects. It should read in the data and use k-fold cross validation to find the optimal lambda as the lowest average MAPE on the k hold out sets. The function should probably return the `lambda_` and the average MAPE.

**AC:**
1. A function that takes a list of possible `lambda_` values, the features and fixed effects we are using, any other inputs that are necessary and run k-fold cross validation to pick the best `lambda_` value to use
2. Unit tests

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2418


jchaskell (dismissed):
This is a really great start on the cross validation function! A few comments / changes - some are things to make the function more general.

src/elexmodel/client.py
```diff
@@ -222,6 +301,20 @@ def get_estimates(
         )
         unexpected_units = data.get_unexpected_units(percent_reporting_threshold, aggregates)
 
+        # get new lambda value from config
+        new_lambda_, avg_MAPE = self.compute_lambda(preprocessed_data, model_settings, lambda_, features, estimands)
```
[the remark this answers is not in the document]
rishasurana: Length of the lambdas list makes the most sense to me!

src/elexmodel/client.py
```diff
@@ -118,6 +124,79 @@ def get_all_conformalization_data_agg(self):
...
+        K=3,
+    ):
+        if len(features) == 0 or len(estimands) == 0 or len(possible_lambda_values) == 0:
+            return 0, 0
+
+        average_MAPE_sum = 0
+        counter = 0
+        best_lambda = None
+        best_MAPE = float("inf")
+        estimand = estimands[0]
```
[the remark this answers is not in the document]
rishasurana: In the most recent revision, I passed our current list of estimands directly from model_settings for ease of access, and then added a parameter to allow us to choose which estimand we want. This allows the current test structure to remain the same and all the modifications are only done while calling fit_model and not in the BaseElectionModel itself. Does this work?

jchaskell (commented):
This looks really good now! I think there are a few things left from your last change that maybe we can delete or revert back? Otherwise, let's see what @lennybronner thinks.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -6,6 +6,8 @@
 import cvxpy
 import numpy as np
 from elexsolver.QuantileRegressionSolver import QuantileRegressionSolver
+from sklearn.metrics import mean_absolute_percentage_error
+from sklearn.model_selection import KFold
```
lennybronner, line 10: I'm curious if there is a way to do this without using scikit-learn, which is a pretty large library and would add significantly to the overhead of this repo. For `mean_absolute_percentage_error` we can use `elexmodels.math_utils.compute_error(true, pred, type_='mape')`. `KFold` might be a bit harder to implement, but still worth it I think.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+
+    def compute_lambda(
+        self,
+        reporting_units,
+        possible_lambda_values=[],
+        estimand="",
+        K=3,
+    ):
+        if (
+            len(self.features) == 0
```
lennybronner, line 279: curious why we are checking this?

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+            or len(possible_lambda_values) == 0
+            or estimand == ""
+            or not (f"{estimand}" in reporting_units)
```
lennybronner, line 282: also curious about these two checks

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+        ):
+            return 0, 0
+
+        MAPE_arr = np.full_like(possible_lambda_values, 0)  # array of MAPE sums for each lambda
```
lennybronner, line 286: this is totally fine as is, but `zeros_like` does this without having to set the value.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+
+        # get the data section indexes that we will be training/testing on
+        divisor = 0
+
+        for train_index, test_index in kfold.split(reporting_units):
+            divisor += 1
+            train = reporting_units.iloc[train_index]
+            test = reporting_units.iloc[test_index]
+            # loop through each lambda
+            index = 0
```
lennybronner, line 297: if you wrap the for loop in an enumerate you can get the index for free. ie:
```
for index, (train_index, test_index) in enumerate(kfold.split(reporting_units)):
```

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+            for lam in possible_lambda_values:
```
lennybronner, line 298: I would switch the order, first iterating over the lambdas and then the splits, since you can do the average after every outer iteration, you don't need to keep track of the divisor anymore.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+                # build model with custom lambda
+                self.lambda_ = lam
+                unit_predictions = self.get_unit_predictions(train, test, estimand)
+                MAPE = mean_absolute_percentage_error(test[estimand].values, unit_predictions.values)
+                MAPE_arr[index] += MAPE
+                index += 1
+
+        # determine average and best
+        MAPE_arr_avg = [value / divisor for value in MAPE_arr]
```
lennybronner, line 307: You don't need to iterate over the values in a numpy array, you can divide like this:
```
MAPE_arr_avg = MAPE_arr / divisor
```
This does the iteration within numpy, which is often super optimized.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +267,46 @@ def get_coefficients(self):
...
+        best_MAPE_index = MAPE_arr_avg.index(min(MAPE_arr_avg))
```
lennybronner, line 308: `np.argmin` might be the more explict version of this

lennybronner (commented):
This looks really great! I left a few small numpy suggestions. I am also curious if we can avoid using scikit-learn, since this is a pretty large package and will make installing the model quite a bit slower.

src/elexmodel/models/BaseElectionModel.py
rishasurana: 1. Removed `or estimand == ""`.
2. The `or not (f"{estimand}" in reporting_units)` check is neccessary in the case that the estimand is not in the reporting_units. This seems to only happen when the estimand is "turnout." In that case, should I change this check to check if the estimand is "turnout" and go from there?

src/elexmodel/client.py
```diff
@@ -91,9 +91,11 @@ def _check_input_parameters(
...
         if not isinstance(robust, bool):
             raise ValueError("robust is not valid. Has to be a boolean.")
-        if not isinstance(lambda_, (float, int)):
-            raise ValueError("lambda is not valid. It has to be numeric.")
-        if lambda_ < 0:
+        if not isinstance(lambda_, list):
+            raise ValueError("lambda is not valid. It has to be a list.")
+        if not all(isinstance(e, (int, float)) for e in lambda_) and len(lambda_) > 0:
+            raise ValueError("lambda is not valid. It has to be a list of numbers.")
+        if len(lambda_) < 0:
```
lennybronner, line 98: I'm not 100% sure I understand how this could happen?

src/elexmodel/client.py
```diff
@@ -91,9 +91,11 @@ def _check_input_parameters(
             raise ValueError("beta is not valid. Has to be either an integer or a float.")
         if not isinstance(robust, bool):
             raise ValueError("robust is not valid. Has to be a boolean.")
-        if not isinstance(lambda_, (float, int)):
-            raise ValueError("lambda is not valid. It has to be numeric.")
-        if lambda_ < 0:
+        if not isinstance(lambda_, list):
+            raise ValueError("lambda is not valid. It has to be a list.")
+        if not all(isinstance(e, (int, float)) for e in lambda_) and len(lambda_) > 0:
```
lennybronner, line 96: I think turning this around makes a bit more sense:
```
if len(lambda_) > 0 and not all(isinstance(e, (int, float)) for e in lambda_):
```
Since if `len(lambda_) < 0` this will short circuit and the second part will never get checked.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,36 @@ def get_coefficients(self):
...
         return self.features_to_coefficients
+
+    def compute_lambda(
+        self,
+        reporting_units,
+        possible_lambda_values=[],
+        estimand="",
+        K=3,
+    ):
+        if len(possible_lambda_values) == 0:
```
lennybronner, line 277: if `possible_lambda_values` just has one element in it, then I don't think we would want to run `compute_lambda` either, right? Since we know what the optimal lambda would be from that set.

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,36 @@ def get_coefficients(self):
...
+            return 0, 0
+
+        MAPE_arr = np.zeros_like(possible_lambda_values)  # array of MAPE sums for each lambda
+
+        # loop through each lambda
+        for loc, lam in enumerate(possible_lambda_values):
+            self.lambda_ = lam
```
lennybronner, line 284: we aren't using `self.lambda_` anymore, right?

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,36 @@ def get_coefficients(self):
         These coefficients are for the point prediciton only, not for the lower or upper intervals models.
         """
         return self.features_to_coefficients
+
+    def compute_lambda(
```
lennybronner, line 270: this is truly the tiniest thing, but do you mind changing the name of this function to `find_optimal_lambda`

src/elexmodel/utils/math_utils.py
```diff
@@ -93,3 +93,20 @@ def compute_mean_pi_length(lower, upper, pred):
     """
     # we add 1 since pred can be literally zero
     return np.mean((upper - lower) / (pred + 1)).round(decimals=2)
+
+
+def get_kfold_splits(data, n_splits):
```
lennybronner, line 98: do you mind adding some comments to this function and also unit tests for it?

src/elexmodel/utils/math_utils.py
```diff
@@ -93,3 +93,20 @@ def compute_mean_pi_length(lower, upper, pred):
...
+    splits = []
+    num_samples = len(data)
+    fold_size = num_samples // n_splits
+    remaining_samples = num_samples % n_splits
+    current_index = 0
+
+    for i in range(n_splits):
+        fold_length = fold_size + i if i < remaining_samples else fold_size
+        test_index = range(current_index, current_index + fold_length)
+        train_index = list(set(range(num_samples)) - set(test_index))
```
lennybronner, line 108: I think this may need to be `range(num_samples + 1)` but a unit tests should clear that up should clear that up

src/elexmodel/client.py
rishasurana: I put this check in just in case there's a typo while inputting the lambda values- is that fine? I switched the order of the two checks on that line like you mentioned in a comment below

src/elexmodel/utils/math_utils.py
rishasurana: Based on my unit tests, I left it range(num_samples) but let me know what you think!

src/elexmodel/utils/math_utils.py
```diff
@@ -89,7 +89,37 @@ def compute_frac_within_pi(lower, upper, results):
...
+
+
+def get_kfold_splits(data, n_splits=2):
+    """
+    Computes k-fold splits of given data, depending on the n_splits, for cross-validation,
+    returns a list of paired train and test indexes
+    """
+
+    if n_splits <= 0:
+        raise ZeroDivisionError("n_splits cannot be equal to zero")
```
lennybronner, line 105: great idea adding the errors! But I think they can both be `ValueErrors`

tests/models/test_base_election_model.py
```diff
@@ -300,3 +299,57 @@ def test_get_aggregate_predictions(va_governor_precinct_data):
...
+            "baseline_b": [9, 2, 4, 5],
+            "a": [2, 2, 3, 7],
+            "b": [2, 3, 4, 5],
+        }
+    )
+
+    new_lambda, avg_MAPE = model.compute_lambda(df_X, lambda_, "a")
+
+    assert new_lambda == 0.01
+    assert avg_MAPE == 0.625
```
lennybronner, line 328: kinda curious where you got the 0.625 from?

tests/utils/test_math_utils.py
```diff
@@ -136,3 +136,70 @@ def test_compute_mean_pi_length():
...
+    array = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]]
+    n_splits = 10
+
+    splits = math_utils.get_kfold_splits(array, n_splits)
+
+    assert len(splits) == 10
+    assert len(splits[0][0]) == 3
+
+
+def test_scikit_example():
```
lennybronner, line 197: this would be numpy, I think? also while you're at it, could you add one from a pandas dataframe, since that is what we are actually splitting?

src/elexmodel/utils/math_utils.py
lennybronner: When I pass `data = [1, 2, 3, 4, 5, 6, 7, 8, 9]` and `n_splits=4` then I get this:
```
[([2, 3, 4, 5, 6, 7, 8], range(0, 2)), ([0, 1, 4, 5, 6, 7, 8], range(2, 4)), ([0, 1, 2, 3, 6, 7, 8], range(4, 6)), ([0, 1, 2, 3, 4, 5, 8], range(6, 8))]
```
but subsetting is not inclusive, so if I grab `[2:8]` from `data` I end up with `[3, 4, 5, 6, 7, 8]` instead of `[3, 4, 5, 6, 7, 8, 9]`.  Also I think we'll wat to cast those ranges into lists explicitly (so on line 120) just to be safe.

tests/models/test_base_election_model.py
rishasurana: Initially, I had a bug where the average_MAPE was not consistent with the chosen lambda so I used that check during testing. I can remove it if needed!

src/elexmodel/utils/math_utils.py
lennybronner: I'm not 100% sure what the conclusion of this is

tests/models/test_base_election_model.py
lennybronner: no, it's good to have. but like did you hand compute it?

rishasurana: Yes, checked it by hand for 0.01

lennybronner: amazing! do you mind just adding a comment there that you did that (so we don't get confused about where that 0.625 comes from later on?

src/elexmodel/utils/math_utils.py
rishasurana: Just added a commit that should fix this

rishasurana: Since we are using iloc, I don't think we lose that last number- I added a unit test with this example

lennybronner (commented):
ok, so close. but I also just realized there is something weird happening with the cli. The parameter names don't match, I think (and also you will need to update the default in `cli.py`

rishasurana:
> ok, so close. but I also just realized there is something weird happening with the cli. The parameter names don't match, I think (and also you will need to update the default in `cli.py`

Changed it such that it can be called like this: --lambda='[0.01, 0.05, 0.90, 0.56]' , but the internal variable name is still lambda_. The other alternative is using the click.option() multiple=True parameter (as used with some of the other params) which would require us to pass in values like this: --lambda 0.01 --lambda 0.05 --lambda 0.90 --lambda 0.56

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,37 @@ def get_coefficients(self):
...
+                train = reporting_units.iloc[train_index]
+                test = reporting_units.iloc[test_index]
+                unit_predictions = self.get_unit_predictions(train, test, estimand, lam)
+                MAPE = math_utils.compute_error(
+                    test[f"results_{estimand}"].values, unit_predictions.values, type_="mape"
+                )
+                MAPE_arr[loc] += MAPE
+
+        # determine average and best
+        MAPE_arr_avg = MAPE_arr / len(possible_lambda_values)
```
lennybronner, line 297: I think the average should be diving by K, since that is how many runs we did for each possible lambda value

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,37 @@ def get_coefficients(self):
...
+        possible_lambda_values=[],
+        estimand="",
+        K=3,
+    ):
+        if len(possible_lambda_values) == 0:
+            return 0, None
+        if len(possible_lambda_values) == 1:
+            return possible_lambda_values[0], None
+
+        MAPE_arr = np.zeros_like(possible_lambda_values)  # array of MAPE sums for each lambda input
```
lennybronner, line 282: `zeros_like` returns an array where the `dtype` is `int64` that means that all values will end up being cast an integers (so all values here are zero).

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -264,3 +266,37 @@ def get_coefficients(self):
...
+
+        # loop through each lambda
+        for loc, lam in enumerate(possible_lambda_values):
+            for train_index, test_index in math_utils.get_kfold_splits(reporting_units, K):
+                # build model with custom lambda
+                train = reporting_units.iloc[train_index]
```
lennybronner, line 288: I think we will want to randomize the order of the reporting units before splitting to make sure that our MAPE value isn't dependent on the ordering that we get

lennybronner (commented):
If I run the code with postal code fixed effects I get a nan error:
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=30 --aggregates=postal_code --aggregates=unit --fixed_effects=postal_code --lambda='[1, 10, 100, 1000]'
```

src/elexmodel/models/BaseElectionModel.py
rishasurana: Thanks for catching that! Changed the value- the assertions changed as well (the best average_MAPE is now ~0.83 rather than 0.625), but since we are just dividing by a different number, the MAPE calculations I did by hand are still consistent and the optimal lambda for both tests are still the same

rishasurana: Randomizing the order of reporting units resulted in, often, more than one lambda value having the same average_MAPE regardless of dataset size. The smaller the dataset, the more unique the lambda values. Further, the average MAPE changes consistently so I had to remove the assertions for calculations done by hand.

rishasurana:
> If I run the code with postal code fixed effects I get a nan error:
> 
> ```
> elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=30 --aggregates=postal_code --aggregates=unit --fixed_effects=postal_code --lambda='[1, 10, 100, 1000]'
> ```

When getting unit predictions for each fold `nonreporting_units_features = self.featurizer.featurize_heldout_data(nonreporting_units)`, there is number of `NAN` values (1-2% of the total df) in `nonreporting_units_features` post calling featurize_heldout_data on the test set of each fold. I can't seem to replicate this issue for other fixed-effect/aggregate combinations and it works fine when not splitting the data into test/train sets. Each time I run this command, certain states have NAN values and cause errors more often than others (HI, DE, NJ). Which states cause errors varies each time because the fold data is now randomly selected.

Example runs:
Columns that have NAN values, total NAN values in dataframe
1. ['postal_code_HI'], 303
2. ['postal_code_HI', 'postal_code_NJ', 'postal_code_NV'], 927 
3. ['postal_code_DE', 'postal_code_NJ', 'postal_code_WY'], 930
4. ['postal_code_HI', 'postal_code_WA'], 602
5. ['postal_code_DE', 'postal_code_MD'], 632
6. ['postal_code_DE'], 281

Using `fillna(0.0)` here in the Featurizer class: `new_heldout_data = new_heldout_data.join(missing_expanded_fixed_effects_df)` resolves this issue because it seems like certain states may be inconsistent in certain folds for the smaller test sets and so when we use `.join` it fills them with NAN.

Switching to use `.merge` doesn't work here because there are no common columns to merge on.

rishasurana:
Working on the 2020 election write-up and noticed the same thing I noticed here- that more than one lambda value has the same average_MAPE, even if the lambda values are really far apart. The smaller the dataset, the more unique the avg mape values. As the test dataset gets larger, every lambda value has the same average mape.

rishasurana:
**It appears that lambda has little to no effect on our data.**

Command: 
`elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=50 --aggregates=postal_code --aggregates=unit --lambda='[18347589365782638975893745893653.0, 2.0, 3345683987523849082439027492748237490278420384928394829089032742355.0, 5.0, 9.0, 15.0, 25.0, 37.0, 483475289735820748302742374823742342.0, 1345325206.0, 99.0]'`

Result: 
```
More than one lambda has the lowest average MAPE of: 0.067
[1.834758936578264e+31, 2.0, 3.3456839875238492e+66, 5.0, 9.0, 15.0, 25.0, 37.0, 4.834752897358207e+35, 1345325206.0, 99.0]
```

I've tested large and small lambda values and that still doesn't make a difference in the predictions. For example, let's consider the unit predictions for fold 1:
```
Lambda 1.834758936578264e+31: [421, 2420.0]
Lambda 2.0: [421, 2420.0]
Lambda 3.3456839875238492e+66: [421, 2420.0]
Lambda 5.0: [421, 2420.0]
Lambda 9.0: [421, 2420.0]
Lambda 15.0: [421, 2420.0]
Lambda 25.0: [421, 2420.0]
Lambda 37.0: [421, 2420.0]
Lambda 4.834752897358207e+35: [421, 2420.0]
Lambda 1345325206.0: [421, 2420.0]
Lambda 99.0: [421, 2420.0]
```

The same pattern holds true for the unit predictions of fold 2 and fold 3 (fold 2 != fold 3 ever, but fold 2 + lambda 1 = fold 2 + lambda 2). The fold train/test dataset that is given, consistently produces the same unit predictions across lambda values. So, the fold MAPE values calculated are the same for each lambda value.

I've removed default values and verified that the lambda values are correctly being passed through to the QR when called in find_optimal_lambda() and are being added to the loss function: 

loss function:
```
Sum([6.23963173e-04 1.53951848e-05 8.20251779e-04 ... 7.51431638e-06
 1.22245098e-04 1.22794926e-05] @ (Promote(0.5, (1517,)) @ abs([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370) + Promote(0.0, (1517,)) @ ([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370)), None, False)
```
loss function with lambda:
```
Sum([6.23963173e-04 1.53951848e-05 8.20251779e-04 ... 7.51431638e-06
 1.22245098e-04 1.22794926e-05] @ (Promote(0.5, (1517,)) @ abs([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370) + Promote(0.0, (1517,)) @ ([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370)), None, False) + 1.6786870797807898e+26 @ power(Pnorm(var370[1], 2), 2.0)
```

From what I've read some reasons that this could be happening are that the dataset is already well-behaved, the features are not highly correlated, or the model is not that complex. Adding features also doesn't seem to help this case:

command: 
`elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=50 --aggregates=postal_code --aggregates=unit --lambda='[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000, 99.0]' --features=ethnicity_european --features=ethnicity_east_and_south_asian --features=ethnicity_hispanic_and_portuguese --features=ethnicity_likely_african_american`

output:
```
More than one lambda has the lowest average MAPE of: 0.067
[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000, 99.0]

```
In the unit tests, the test datasets have a similar issue:

```
def test_find_optimal_lambda_tens():
    """
    Test/view computing lambda
    """
    lambda_ = [0.01, 0.1, 1.0, 10.0, 100.0, 1000.0]
    model_settings = {"features": ["b"]}
    model = BaseElectionModel.BaseElectionModel(model_settings)
    df_X = pd.DataFrame(
        {
            "residuals_a": [1, 2, 3, 4],
            "total_voters_a": [4, 2, 9, 5],
            "last_election_results_a": [5, 1, 4, 2],
            "results_a": [5, 2, 8, 4],
            "results_b": [3, 6, 2, 1],
            "baseline_a": [9, 2, 4, 5],
            "baseline_b": [9, 2, 4, 5],
            "a": [2, 2, 3, 7],
            "b": [2, 3, 4, 5],
        }
    )

    new_lambda, avg_MAPE = model.find_optimal_lambda(df_X, lambda_, "a")

    assert new_lambda in [0.01, 10.0]  # more than one lambda value can have the same average MAPE
```
Here, the results oscillate between 
1. More than one lambda has the lowest average MAPE of: 1.0
[0.01, 0.1]
2. More than one lambda has the lowest average MAPE of: 0.417
[10.0, 100.0, 1000.0]
So the best chosen lambda for this test is either 0.01 or 10.0 (since it chooses the first given lambda with the lowest average MAPE). This happens for the other unit tests as well. For such a small dataset, there is a slight bit more variation (2 distinct MAPEs on average for a set of 5 lambdas). For a large dataset such as 2020-11-03_USA_G, there is only 1 distinct MAPE for a set of 5-10 lambdas.

My conclusion is that regularizing our model does not seem to impact the predictions, but I'm not sure if this is a valid conclusion?

rishasurana:
**It appears that lambda has little to no effect on our model predictions... for most cases**

_command:_
`elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=50 --aggregates=postal_code --aggregates=unit --lambda='[18347589365782638975893745893653.0, 2.0, 3345683987523849082439027492748237490278420384928394829089032742355.0, 5.0, 9.0, 15.0, 25.0, 37.0, 483475289735820748302742374823742342.0, 1345325206.0, 99.0]'`
_output:_
```
More than one lambda has the lowest average MAPE of: 0.067
[1.834758936578264e+31, 2.0, 3.3456839875238492e+66, 5.0, 9.0, 15.0, 25.0, 37.0, 4.834752897358207e+35, 1345325206.0, 99.0]
```
I've tested large and small lambda values and that still doesn't make a difference in the predictions. For example, let's consider the unit predictions for fold 1:
```
Lambda 1.834758936578264e+31: [421, 2420.0]
Lambda 2.0: [421, 2420.0]
Lambda 3.3456839875238492e+66: [421, 2420.0]
Lambda 5.0: [421, 2420.0]
Lambda 9.0: [421, 2420.0]
Lambda 15.0: [421, 2420.0]
Lambda 25.0: [421, 2420.0]
Lambda 37.0: [421, 2420.0]
Lambda 4.834752897358207e+35: [421, 2420.0]
Lambda 1345325206.0: [421, 2420.0]
Lambda 99.0: [421, 2420.0]
```
The same pattern holds true for the unit predictions of fold 2 and fold 3 (fold 2 != fold 3 ever, but fold 2 + lambda 1 = fold 2 + lambda 2). The fold train/test dataset that is given, consistently produces the same unit predictions across lambda values. So, the fold MAPE values calculated are the same for each lambda value.

I've removed default values and verified that the lambda values are correctly being passed through to the QR when called in find_optimal_lambda() and are being added to the loss function:
_loss function:_
```
Sum([6.23963173e-04 1.53951848e-05 8.20251779e-04 ... 7.51431638e-06
 1.22245098e-04 1.22794926e-05] @ (Promote(0.5, (1517,)) @ abs([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370) + Promote(0.0, (1517,)) @ ([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370)), None, False)
```
_loss function with lambda:_
```
Sum([6.23963173e-04 1.53951848e-05 8.20251779e-04 ... 7.51431638e-06
 1.22245098e-04 1.22794926e-05] @ (Promote(0.5, (1517,)) @ abs([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370) + Promote(0.0, (1517,)) @ ([0.37310912 0.125      0.15607195 ... 0.53658537 0.13643178 0.35820896] + -[[1.]
 [1.]
 [1.]
 ...
 [1.]
 [1.]
 [1.]] @ var370)), None, False) + 1.6786870797807898e+26 @ power(Pnorm(var370[1], 2), 2.0)
```

From what I've read some reasons that this could be happening are that the dataset is already well-behaved, the features are not highly correlated, or the model is not that complex.

Adding features to the command also doesn't seem to help this case:
_command:_
`elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=50 --aggregates=postal_code --aggregates=unit --lambda='[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000, 99.0]' --features=ethnicity_european --features=ethnicity_east_and_south_asian --features=ethnicity_hispanic_and_portuguese --features=ethnicity_likely_african_american`
_output:_
```
More than one lambda has the lowest average MAPE of: 0.067
[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000, 99.0]
```

Modifying the previous 03_USA_G command that had lots of features with some values closer to zero does nothing:
_command:_
`elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=50 --aggregates=postal_code --aggregates=unit --lambda='[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000, 0, 0.000006]' --features=ethnicity_european --features=ethnicity_east_and_south_asian --features=ethnicity_hispanic_and_portuguese --features=ethnicity_likely_african_american`

_output:_
More than one lambda has the lowest average MAPE of: 0.073
[12.0, 2.0, 33.0, 5.0, 9.0, 15.0, 25.0, 37.0, 234, 1000]


**So far the only test that seems to work as intended is this integration test:**
_command:_
`elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=precinct --aggregates=county_classification --aggregates=postal_code --fixed_effects=county_classification --percent_reporting 10 --features=ethnicity_european --features=ethnicity_hispanic_and_portuguese --lambda='[50, 0.5, 10, 100, 437837, 2, 0]'`

_output:_
avg_MAPE values for each lambda: [0.143 0.137 0.143 0.143 0.143 0.143 0.097]
best: 0

Here, for a large dataset, we are able to produce more variation (3 distinct) in terms of average MAPE. 5 of the 7 given lambdas have the same MAPE. The issue with this test working as intended means that this is less likely to be a logic issue with the code and more likely to be something else.


In the unit tests, the test datasets have a similar issue as before, producing a maximum of 2 unique MAPEs by the end of each run.
```
def test_find_optimal_lambda_tens():
    """
    Test/view computing lambda
    """
    lambda_ = [0.01, 0.1, 1.0, 10.0, 100.0, 1000.0]
    model_settings = {"features": ["b"]}
    model = BaseElectionModel.BaseElectionModel(model_settings)
    df_X = pd.DataFrame(
        {
            "residuals_a": [1, 2, 3, 4],
            "total_voters_a": [4, 2, 9, 5],
            "last_election_results_a": [5, 1, 4, 2],
            "results_a": [5, 2, 8, 4],
            "results_b": [3, 6, 2, 1],
            "baseline_a": [9, 2, 4, 5],
            "baseline_b": [9, 2, 4, 5],
            "a": [2, 2, 3, 7],
            "b": [2, 3, 4, 5],
        }
    )
    new_lambda, avg_MAPE = model.find_optimal_lambda(df_X, lambda_, "a")
    assert new_lambda in [0.01, 10.0]  # more than one lambda value can have the same average MAPE
```
Here, the results oscillate between
1. More than one lambda has the lowest average MAPE of: 1.0
[0.01, 0.1]
2. More than one lambda has the lowest average MAPE of: 0.417
[10.0, 100.0, 1000.0]

So the best chosen lambda for this test is either 0.01 or 10.0 (since it chooses the first given lambda with the lowest average MAPE). This happens for the other 2 unit tests as well. For such a small dataset, there is a slight bit more variation (2 distinct MAPEs on average for a set of 5 lambdas). For a large dataset such as 2020-11-03_USA_G, there is only 1 distinct MAPE for a set of 5-10 lambdas. For 2017-11-07_VA_G, we get a maximum of 3 but only for one of the integration tests. The rest of the integration tests only have 1 distinct MAPE (all the same).

My conclusion is that regularizing our model does not seem to impact the predictions in most cases, but I'm not sure if this is a valid conclusion?

jchaskell:
Lenny should definitely weigh in, but a couple of thoughts:
* Regularization only happens when we have features and/or fixed effects, so I don't think it's unexpected to get the same value in cases where we aren't specifying either.
* I may be remembering this wrong, but I think generally lambda is set between 0 and 1 (though that may be for a different type of regression than what we're using). So it may not be that surprising that we're getting the same values above 1 in a lot of cases. We may want to test more values in that range or around the minimums you are seeing to see if we can identify the actual best lambda.
* But to your point, that even when we're getting different error rates, the actual values seem similar - it's not super clear to me what a different of `0.01` in MAPE means in real terms. I think I'd want to see the actual predictions for a couple of close states with the two lambdas to understand that, though maybe Lenny has another idea of how to look at differences.

PR #58, opened 2023-06-27
ELEX-2286-winsorize

rishasurana:
## Description
In the Gaussian model, there are sometimes outlier points in the conformalization set that skew confidence intervals. We have experimented with some ways of downweighting or cutting out the outliers.

We add a model parameter that allows user to sub in different methods of dealing with these data outliers. “Windsorizing” is one of them. We add a model parameter for the user when using the Gaussian model. They can windsorize, or use all the conformalization data as usual. There may be other methods we want to add as well. This parameter should be passed in with all the other model settings.

## Jira Ticket
[https://arcpublishing.atlassian.net/browse/ELEX-2286
](https://arcpublishing.atlassian.net/browse/ELEX-2286
)

## Notes
This PR sets the winsorize parameter as a boolean to allow the user to choose whether they want to call `robust_sample_std` (with winsorizing built in) or just `sample_std` (no winsorizing) when calling `boot_sigma`.

This PR is a new iteration of PR#26 in the model repo to reduce merge conflicts.

## Test Steps
tox


src/elexmodel/distributions/GaussianModel.py
```diff
@@ -105,8 +107,10 @@ def to_aggregate(x):
                                 x[f"last_election_results_{estimand}"] / np.sum(x[f"last_election_results_{estimand}"])
                             ).to_numpy(),
                         ),
-                        "sigma_lower_bound": beta * math_utils.boot_sigma(x.lower_bounds.values, conf=(3 + alpha) / 4),
-                        "sigma_upper_bound": beta * math_utils.boot_sigma(x.upper_bounds.values, conf=(3 + alpha) / 4),
+                        "sigma_lower_bound": beta
+                        * math_utils.boot_sigma(x.lower_bounds.values, conf=(3 + alpha) / 4, winsorize=winsorize),
```
lennybronner, line 111: I'd prefer if we passed winsorize as a parameter to `boot_sigma` without having to add it as a column in the dataframe

src/elexmodel/distributions/GaussianModel.py
```diff
@@ -70,7 +72,7 @@ def _get_n_units_per_group(self, conformalization_data, nonreporting_units, aggr
             .fillna({"n": 0})
         )
 
-    def _fit(self, conformalization_data, estimand, aggregate, alpha, beta):
+    def _fit(self, conformalization_data, estimand, aggregate, alpha, beta, winsorize):
```
lennybronner, line 75: I just realize we need don't need to pass `beta` or `winsorize` into `fit` or `_fit` because they are passed into the `GaussianModel` constructor as part of `model_settings`, right?

src/elexmodel/utils/file_utils.py
```diff
@@ -3,6 +3,10 @@
 import pathlib
 from io import StringIO
 
+from dotenv import find_dotenv, load_dotenv
+
+load_dotenv(find_dotenv())
+
```
lennybronner, line 9: I think this can be removed, right?

src/elexmodel/utils/math_utils.py
```diff
@@ -55,11 +65,23 @@ def weighted_median(x, weights):
...
+def boot_sigma(data, conf, num_iterations=10000, winsorize=winsorize):
     """
     Bootstrap standard deviation.
     """
     # we use upper bound of confidence interval for more robustness
+
+    if winsorize:
+        return bootstrap(
+            data.reshape(1, -1), robust_sample_std, confidence_level=conf, method="basic", n_resamples=num_iterations
+        ).confidence_interval.high
```
lennybronner, line 84: I think this can be simplified a bit by defining a variable called `std_func` which is dependent on whether `winsorize` is true or not, which then gets passed into `bootstrap` either way

src/elexmodel/distributions/GaussianModel.py
rishasurana: Could you clarify where I'm adding winsorize to a dataframe?

src/elexmodel/cli.py
```diff
@@ -39,6 +39,7 @@
     type=click.Choice(["county", "precinct", "county-district", "precinct-district"]),
 )
 @click.option("--beta", "beta", default=1, type=int, help="manually add variance to Gaussian model")
+@click.option("--winsorize", "winsorize", default=False, type=bool, help="reduce outliers in the Gaussian model")
```
lennybronner, line 42: I think we want to make this a flag (rather than an explicit parameter). Take a look at what robust looks like just below. The difference is:
```
elexmodel 2021-01-05_GA_G --office_id=S_precinct --estimands=dem --geographic_unit_type=precinct --pi_method=gaussian --percent_reporting=5 --winsorize=True
```
vs. 
```
elexmodel 2021-01-05_GA_G --office_id=S_precinct --estimands=dem --geographic_unit_type=precinct --pi_method=gaussian --percent_reporting=5 --winsorize
```

tests/models/test_gaussian_election_model.py
```diff
@@ -1,23 +1,6 @@
 from elexmodel.models import GaussianElectionModel
 
 
-def test_instantiation():
```
lennybronner, line 4: curious what is happening here

tests/test_client.py
```diff
@@ -224,6 +234,29 @@ def test_check_input_parameters_beta(model_client, va_governor_config):
...
             pi_method,
             "bad_beta",
+            winsorize,
+            robust,
+            lambda_,
+            handle_unreporting,
+        )
+
+
+def test_check_input_parameters_winsorize(model_client, va_config):
```
lennybronner, line 244: This unit test is failing for me. Can you try to merge develop into this branch?

lennybronner (commented):
This looks great! 
I do think it would be really useful to add a unit test that runs the model with winzorization against one without and checks that the prediction intervals with winsorization are smaller than those without.

tests/models/test_gaussian_election_model.py
rishasurana: This check was one I wrote from the first iteration when we were passsing beta and winsorize into the model directly through fit. When we decided to pass them through model settings instead, it was no longer needed

tests/distributions/test_gaussian_model.py
```diff
@@ -129,28 +129,33 @@ def test_fit():
...
     alpha = 0.9
-    beta = 1
     estimand = "turnout"
     model_settings = {
         "election_id": "2017-11-07_VA_G",
         "office": "G",
         "geographic_unit_type": "county",
         "save_conformalization": False,
+        "beta": 1,
+        "winsorize": True,
```
lennybronner, line 139: I think this should be False now, since we have a separate `test_winsorize` unit test?

lennybronner (approved):

PR #59, opened 2023-06-29
ELEX-2763-estimandizer

rishasurana:
## Description
At the moment we are always estimating something that is passed in directly to get_estimates (and something that is directly in the preprocessed_data), this includes dem_votes, gop_votes and total_votes. We want to generalize this process by letting us generate estimands on the fly given the live data or preprocessed data.

Likely, this means creating a class or a function (similar to the Featurizer) that given a data handler (ie. PreprocessedDataHandler, LiveDataHandler or CombinedDataHandler )and a list of estimands, will generate those features in the dataframe. 

Since dem_votes, gop_votes and total_votes are already included in the data, this estimand generator should do nothing (or over-write those fields). But it should let us create more complex estimands if we want (ie. margin)

## Jira Ticket
`https://arcpublishing.atlassian.net/browse/ELEX-2763`

## Test Steps
1. test_estimandizer.py
2. test_client.py
3. Test to run client directly with a pre-written function: `elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}" `
4. Test to run client directly with a given function: `elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{None:'Test.hello'}"`


src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,153 @@
...
+            "ethnicity_voting_patterns": [self.calculate_party_vote_share, self.examine_ethnicity_voting_patterns],
+            "income_impact": [
+                self.calculate_party_vote_share,
+                self.standardize_median_household_income,
+                self.explore_income_impact,
+            ],
+            "candidate": [self.candidate],
+        }
+
+    def check_estimand(self, estimand):
```
jchaskell, line 30: I think we want to check the data and not have a pre-specified list here

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,153 @@
...
+        return len(missing_columns) == 0
+
+    def predict_estimands(self, estimand):
+        """
+        Predict which estimands can be formed given a dataset and a list of estimands we would like to create
+        """
+        if estimand not in self.transformation_map:
+            raise ValueError(f"Estimand '{estimand}' is not supported.")
+        self.transformations = self.transformation_map[estimand]
+        """
```
jchaskell, line 53: this shouldn't be commented out right?

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,153 @@
...
+        return True
+
+    def check_input_columns(self, columns):
+        """
+        Check that input columns contain all neccessary values for a calculation
+        """
+        missing_columns = [col for col in columns if col not in self.data_handler.data.columns]
+        return len(missing_columns) == 0
+
+    def predict_estimands(self, estimand):
```
jchaskell, line 46: Can we change the name of this method? Maybe something like `verify_estimand` or `is_estimand_possible()`? Just cuz we're not really predicting anything.

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Estimandizer. Generate estimands explicitly.
+    """
+
+    def __init__(self, data_handler, estimands):
+        self.data_handler = data_handler
+        self.estimands = estimands
+        self.transformations = []
+        self.transformation_map = {
+            "margin": [self.calculate_margin],
```
jchaskell, line 14: Are these functions Lenny gave you to include?

jchaskell (changes requested):
In general, I really like this approach! Two comments/questions:

1. Maybe this goes against what Lenny said but I think it would be nice to be able to pass in any function as an estimand. We should still keep some defaults like what are in the transformation map but also be able to pass a new one in. Does that make sense?

2. Did he say anything about integrating this into the model code? Or does he just want it created so he can use it when he re-writes things?

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: On the Jira ticket it says that "since dem_votes, gop_votes and total_votes are already included in the data, this estimand generator should do nothing (or over-write those fields)" so that's why I had a pre-specified list. Should I switch this so that we have a parameter of estimands that we want to check, inclusive of these three?

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: Lenny mentioned margin, percent of votes, and candidate name vote / two party vote. I added a few more that I could come up with

rishasurana:
> In general, I really like this approach! Two comments/questions:
> 
> 1. Maybe this goes against what Lenny said but I think it would be nice to be able to pass in any function as an estimand. We should still keep some defaults like what are in the transformation map but also be able to pass a new one in. Does that make sense?
> 2. Did he say anything about integrating this into the model code? Or does he just want it created so he can use it when he re-writes things?

1. Added the capability to pass in a dict of {estimand: function}
2. He didn't mention this and I don't see it on the Jira ticket, but I will ask him when he comes back!

rishasurana:
> In general, I really like this approach! Two comments/questions:
> 
> 1. Maybe this goes against what Lenny said but I think it would be nice to be able to pass in any function as an estimand. We should still keep some defaults like what are in the transformation map but also be able to pass a new one in. Does that make sense?
> 2. Did he say anything about integrating this into the model code? Or does he just want it created so he can use it when he re-writes things?

1. Added the capability to pass in a dict of {estimand: function}
2. He didn't mention this and I don't see it on the Jira ticket, but I will ask him when he comes back!

src/elexmodel/handlers/data/Estimandizer.py
jchaskell: My thinking was that we could check the data and see if the name of the estimand, i.e. `dem_votes`, exists in the data and if it does, do not create it. Does that make sense?

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -36,47 +37,62 @@ def check_estimand(self, estimand):
...
-            [col for transform in self.transformations for col in transform.__code__.co_varnames[1:]]
+            [col for transform in self.transformations for col in transform.__code__.co_varnames[1:]], election_id
         ):
             return []
-        """
+
         return self.transformations
 
-    def create_estimand(self, estimand):
+    def create_estimand(self, estimand, given_function):
```
jchaskell, line 68:
```suggestion
    def create_estimand(self, estimand = None, given_function = None):
```
I think we want to default both of these to None? Or maybe just one of them.

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -36,47 +37,62 @@ def check_estimand(self, estimand):
...
         """
-        if estimand in self.transformation_map:
-            if self.transformation_map[estimand][0] in self.transformations:
-                transformation_func = self.transformations[
-                    self.transformations.index(self.transformation_map[estimand][0])
-                ]
-                transformation_func()
-
-    def generate_estimands(self):
+        if estimand is None:
```
jchaskell, line 72: Can you add a check that one of the two (estimand and given_function) is not None?

jchaskell (dismissed):
Couple of small additional comments. 

Lenny is going to be out for more of next week than I thought but he said he can do some code reviews so we can have him take a look.

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: I added a function at the start of generate_estimands that creates dem_votes, gop_votes and total_votes in a G election if they do not exist. If the data handler is pre-processed and already has the result data (results_dem, results_gop, results_total), it will use that. Otherwise, if the data handler is combined, I have it set such that it creates empty columns with these names. Is this what you mean?

lennybronner (commented):
I really like how this is set up! A few things:
1. We'll want to keep this to the outcome variables only (so really only treating dem/gop/turnout/candidates), the features (like race, income etc.) would be treated in the featurizer.
2. Also for now, let's keep this to dem/gop/turnout/candidate votes and dem/gop/candidate shares (we can add margin and turnout factor later as part of [this](https://arcpublishing.atlassian.net/browse/ELEX-2767) ticket.
3. Do you mind adding how this is going to get called in `client.py` also?

README.md
```diff
@@ -67,6 +67,7 @@ Parameters for the CLI tool:
 | robust               | flag    | flag for larger set of prediction intervals in the nonparametric case |
 | save_output          | list    | `results`, `data`, `config` |
 | unexpected_units     | int     | number of unexpected units to simulate; only used for testing and does not work with historical run |
+| unexpected_units     | dict     | mapping of estimand names to None or pre-built functions |
```
lennybronner, line 70: I think the parameter name here needs to be updated

src/elexmodel/client.py
```diff
@@ -229,6 +235,10 @@ def get_estimates(
             handle_unreporting=handle_unreporting,
         )
 
+        if new_estimands:
```
lennybronner, line 238: Since the `new_estimands` default is an empty dictionary, this means that if I don't specify a `new_estimands` the Estimandizer never gets called, right? but what if I want the estimand to be something for which we have a pre-defined function like party vote share?

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,111 @@
...
+                    transformation_func = self.transformations[
+                        self.transformations.index(self.transformation_map[estimand][0])
+                    ]
+                    transformation_func()
+
+    def generate_estimands(self):
+        """
+        Main function to generate estimands
+        """
+        if self.office == "G":
```
lennybronner, line 70: why is this specific to a governor election?

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,111 @@
...
+    def check_input_columns(self, columns):
+        """
+        Check that input columns contain all neccessary values for a calculation
+        """
+        missing_columns = []
+        if self.office == "G":
+            missing_columns = [col for col in columns if col not in self.data_handler.data.columns]
+        elif self.office == "P":
+            missing_columns = [
+                col for col in columns if col not in self.data_handler.data[self.data_handler.election_id]
```
lennybronner, line 33: why is there a difference in governor and presidential elections?

src/elexmodel/cli.py
```diff
@@ -66,6 +66,13 @@
     help="options: results, data, config",
 )
 @click.option("--handle_unreporting", "handle_unreporting", default="drop", type=click.Choice(["drop", "zero"]))
+@click.option(
+    "--new_estimands",
```
lennybronner, line 70: I would slightly prefer if we called this `estimand_fns` or something or `estimand_fn_dict`

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,111 @@
...
+    def verify_estimand(self, estimand):
+        """
+        Verify which estimands can be formed given a dataset and a list of estimands we would like to create
+        """
+        if estimand not in self.transformation_map:
+            raise ValueError(f"Estimand '{estimand}' is not supported.")
+        self.transformations = self.transformation_map[estimand]
+
+        if not self.check_input_columns(
+            [col for transform in self.transformations for col in transform.__code__.co_varnames[1:]]
```
lennybronner, line 46: do you mind adding a comment for this section?

src/elexmodel/client.py
rishasurana: Yes, if you don't specify a new estimand it will never get called.

If you want to call a predefined function, for instance "party_vote_share" you just need to pass in {"party_vote_share": None} since the generate function checks that if the value is None, then the estimand must be a pre-written function. If you make a new function and want to pass that as well, you can do {"party_vote_share": None, "new_function": [pass in new function here]} and then it will run that function.

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: I meant to look at election_type since they have different input structures (general vs primary)- fixed.

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: Same as above, I meant to look at election_type since they have different input structures (general vs primary)- fixed.

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,113 @@
...
+    def generate_estimands(self):
+        """
+        Main function to generate estimands
+        """
+        if self.election_type == "G":
+            self.pre_check_estimands()
+
+        for key, value in self.estimand_fns.items():
+            if value is None:  # Option 1: Pass in a list of estimands we want to build from a pre-set list
+                self.verify_estimand(key)
```
lennybronner, line 77: how is `verify_estimand` related to `pre_check_estimands()`?

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -0,0 +1,113 @@
...
+    def check_input_columns(self, columns):
+        """
+        Check that input columns contain all neccessary values for a calculation
+        """
+        missing_columns = []
+        if self.election_type == "G":
+            missing_columns = [col for col in columns if col not in self.data_handler.data.columns]
+        elif self.election_type == "P":
+            missing_columns = [
+                col for col in columns if col not in self.data_handler.data[self.data_handler.election_id]
```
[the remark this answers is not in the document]
lennybronner: Also, in the case of a primary the candidates function is the standard not `dem`, `gop` etc.

lennybronner (commented):
Getting an error when running with:
```
 elexmodel 2020-11-03_USA_G --estimands=dem --office_id=P --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}"
```

but also, when I change the estimand to `party_vote_share_dem` this gives me an error later:
```
elexmodel 2017-11-07_VA_G --estimands=party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}"
```
I think we need to use the estimandizer in the mocklivedatahandler too, right?

src/elexmodel/handlers/data/Estimandizer.py
rishasurana: `pre-check-estimands` just looks for and creates the standard values before any complex estimands are created, while `verify_estimand` is called to check if a given complex pre-written estimand is possible. these pre-written estimands usually depend on the existence of the standard estimands

rishasurana:
I've pushed some updates to how the class works/the test cases to ignore election type and added some more comments. 

For:` elexmodel 2017-11-07_VA_G --estimands=dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{None:'Test.hello'}" `

Given a class like this:
```
class Test:
    def __init__():
          print("INIT TEST")
    def hello(obj):
            print("HELLO WORLD")
            obj.data_handler.data["TESTING"] = None
```
And adding the corresponding input statement at the top of Estimandizer.py
`from elexmodel.handlers.data.Test import Test
`
You can input functions from inside and outside the Estimandizer class.


The next steps for this class would be to integrate the MockLiveDataHandler and add functionality where the estimand given does not have to exist prioir to calling the estimandizer so that we can successfully run this: 

`elexmodel 2017-11-07_VA_G --estimands=party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}"`

Currently, based on the way cli.py is set up, as opposed to client.py, initially setting the estimand to "party_vote_share_dem" which does not exist yet, rather than just "dem," will result in an error in LiveData.py (around line 96).

dmnapolitano:
Hi @lennybronner , something mysterious is happening with this branch 🤔 

I merged in `develop` and now the `test_winsorize_intervals` unit test in `tests/test_client.py` keeps failing.  But what's frustrating is, every time it fails, it fails with a different pair of values and in a different spot.  Sometimes the first assert will fail, sometimes that one will pass and the second assert will fail.

So, there's definitely some randomness happening somewhere but I can't figure out where.  In the unit test, the data in the `data_handler` is being shuffled, and I just added a `seed` which you can see here.  But the test is still failing in the same way and I have no idea why 😞

dmnapolitano:
> Getting an error when running with:
> 
> ```
>  elexmodel 2020-11-03_USA_G --estimands=dem --office_id=P --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}"
> ```
> 
> but also, when I change the estimand to `party_vote_share_dem` this gives me an error later:
> 
> ```
> elexmodel 2017-11-07_VA_G --estimands=party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'party_vote_share':None}"
> ```
> 
> I think we need to use the estimandizer in the mocklivedatahandler too, right?

Hi!  Just wanted to follow-up on this as I just encountered it, too.  I find the current way this is set up somewhat unintuitive.  I expect to be able to do:

```
elexmodel 2017-11-07_VA_G --estimands="['dem', 'party_vote_share_dem']" --office_id=G --geographic_unit_type=county --percent_reporting 50
```

And:
* `dem` is known, it's defined in the input data, so we keep doing what we've always been doing;
* `party_vote_share_dem` is not in the input data, so we check to see if we have a function that can create `party_vote_share_dem` and if not, throw an `Exception` saying we need to define said function.

But also:

```
elexmodel 2017-11-07_VA_G --estimands="['dem', 'party_vote_share_dem', 'my_new_estimand']" --office_id=G --geographic_unit_type=county --percent_reporting 50 --estimand_fns="{'my_new_estimand' : ...}"
```

Now, the "..." part is where things get tricky.  Right now, the `Estimandizer` class is doing an `eval()` of whatever "..." is, which is really scary TBH.  It would be safer to replace this with [`SymPy`](https://docs.sympy.org/latest/index.html), or not allow users to define functions like this as strings on the command line.  Instead, one actually has to use the model via API where they can define a function programmatically and pass that function in to their `ModelClient` object.

Thoughts? 🤔

dmnapolitano:
> Hi @lennybronner , something mysterious is happening with this branch 🤔
> 
> I merged in `develop` and now the `test_winsorize_intervals` unit test in `tests/test_client.py` keeps failing. But what's frustrating is, every time it fails, it fails with a different pair of values and in a different spot. Sometimes the first assert will fail, sometimes that one will pass and the second assert will fail.
> 
> So, there's definitely some randomness happening somewhere but I can't figure out where. In the unit test, the data in the `data_handler` is being shuffled, and I just added a `seed` which you can see here. But the test is still failing in the same way and I have no idea why 😞

Update on this: so I reviewed this branch vs. `develop` and vs. `main`, and I couldn't really see any unexpected differences, except, I noticed that `test_estimandizer_input()` was added immediately after `test_winsorize_intervals()`.  So I'm thinking Risha may have added in some formatting around `test_winsorize_intervals()` that lead git to believe that function is a new-ish addition in this `estimandizer` branch.  Anyway, I went ahead and removed `test_winsorize_intervals()` and the unit tests are passing here.  Hopefully that settles that 🤔 🤞🏻

dmnapolitano:
Hi again! 👋🏻 

Alright, so based on our conversation yesterday and the comments I left above, I basically re-wrote this 😅   My thinking here was:

1. The `Estimandizer` should become the one-stop shop for all things estimand-related.  Right now it's just checking that estimands and their related columns exist and creating new ones as specified, but in the future, this could be the central place where we do things like check the bounds/validity of estimands, check for/impute missing values, etc. etc. so that it's standardized across all `DataHandler`s.
2. Speaking of the `DataHandler` classes, since `DataHandler` objects have `data` and data has estimands, a `DataHandler` has-a `Estimandizer`.  So you'll see only `Estimandizer` objects existing within `DataHandler` objects and some refactoring to, again, have `Estimandizer` handle common estimand-related functionality rather than duplicate it across `DataHandler`s (as much as possible 😅 ).

I hope this makes sense and meets what I think it the goal of this work 😅  Thanks and I'm looking forward to your feedback here 🎉

dmnapolitano:
> Hang on, sorry, I missed a spot where estimands are being checked for validity 🤔

Alright, this is ready now (I think) 😅 🙌🏻

tests/test_client.py
```diff
@@ -759,3 +766,38 @@ def test_conformalization_data(model_client, va_governor_county_data, va_config)
     assert isinstance(conform_unit[0.9]["turnout"][1], pd.DataFrame)
     assert conform_agg[0.9]["turnout"][0] is None
     assert isinstance(conform_agg[0.9]["turnout"][1], pd.DataFrame)
+
+
+def test_estimandizer_input(model_client, va_governor_county_data, va_config):
```
lennybronner, line 771: not sure I quite understand what this test is for

tests/test_client.py
```diff
@@ -759,3 +766,38 @@ def test_conformalization_data(model_client, va_governor_county_data, va_config)
...
+            estimands,
+            prediction_intervals,
+            percent_reporting_threshold,
+            geographic_unit_type,
+            raw_config=va_config,
+            preprocessed_data=preprocessed_data,
+            save_output=[],
+        )
+    except KeyError:
+        pytest.raises("Error with clinet input for estimandizer")
```
lennybronner, line 803:
```suggestion
        pytest.raises("Error with client input for estimandizer")
```

src/elexmodel/client.py
```diff
@@ -56,21 +56,27 @@ def _check_input_parameters(
...
-        valid_estimands = config_handler.get_estimands(office)
-        for estimand in estimands:
-            if estimand not in valid_estimands:
-                raise ValueError(f"Estimand: '{estimand}' is not valid. Please check config")
+
+        estimands_in_config = config_handler.get_estimands(office)
+        extra_estimands = set(estimands).difference(set(estimands_in_config))
+        if len(extra_estimands) > 0:
+            # this is ok; later they'll all be created by the Estimandizer
+            LOG.info("Found additional estimands that were not specified in the config file: %s", extra_estimands)
```
lennybronner, line 64: if the user didn't create a new estimand function but just input something random, then this would pass, right?

lennybronner (commented):
This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?

Also, with a historical election I am getting this error:
```
 elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
 ```

File: src/elexmodel/client.py (hunk 17 above)
Comment (dmnapolitano) [reply src/elexmodel/client.py:64]: Yes and no.  Eventually you'll get a `KeyError` when the "something_random" estimand's function can't be found.  I figured a warning here would be helpful for both (a) you have an estimand that you didn't specify in your config and (b) if that `KeyError` does happen, here's an indication as to why 😄

**Comment by dmnapolitano** [comment]
> This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> 
> Also, with a historical election I am getting this error:
> 
> ```
> KeyError: "['results_party_vote_share_dem'] not in index"
> ```
> 
> here is an example
> 
> ```
>  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> ```

Thanks!  Ok, so I just pushed changes that should get the `--historical` to work properly.  Sorry about that and thanks for providing that example 😅 🎉 

As for your other questions:

1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔

File: src/elexmodel/client.py (hunk 17 above)
Comment (lennybronner) [reply src/elexmodel/client.py:64]: that sounds good, thanks!

**Comment by lennybronner** [comment]
> > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > Also, with a historical election I am getting this error:
> > ```
> > KeyError: "['results_party_vote_share_dem'] not in index"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > here is an example
> > ```
> >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > ```
> 
> Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> 
> As for your other questions:
> 
> 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔

Thank you for fixing the historical run!

It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test). 

re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums

**Comment by dmnapolitano** [comment]
> > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > Also, with a historical election I am getting this error:
> > > ```
> > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > here is an example
> > > ```
> > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > ```
> > 
> > 
> > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > As for your other questions:
> > 
> > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> 
> Thank you for fixing the historical run!
> 
> It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> 
> re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums

Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉  Please try it out and let me know what you think.  As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔  Open to suggestions, of course.

**Comment by lennybronner** [comment]
> > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > Also, with a historical election I am getting this error:
> > > > ```
> > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > here is an example
> > > > ```
> > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > ```
> > > 
> > > 
> > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > As for your other questions:
> > > 
> > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > 
> > 
> > Thank you for fixing the historical run!
> > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> 
> Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.

This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.

**Comment by dmnapolitano** [comment]
> > > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > > Also, with a historical election I am getting this error:
> > > > > ```
> > > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > here is an example
> > > > > ```
> > > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > > ```
> > > > 
> > > > 
> > > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > > As for your other questions:
> > > > 
> > > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > > 
> > > 
> > > Thank you for fixing the historical run!
> > > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> > 
> > 
> > Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.
> 
> This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.

Got it, I see, thanks.  Ok, I (a) rolled back the changes I had made to those unit tests, and (b) now if a `baseline_` column was added during the `CombinedDataHandler` estimandization process, they're now deleted.  Let me know what you think 😄

**Comment by lennybronner** [comment]
> > > > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > > > Also, with a historical election I am getting this error:
> > > > > > ```
> > > > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > > > ```
> > > > > > 
> > > > > > 
> > > > > >     
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >     
> > > > > >   
> > > > > > here is an example
> > > > > > ```
> > > > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > > > ```
> > > > > 
> > > > > 
> > > > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > > > As for your other questions:
> > > > > 
> > > > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > > > 
> > > > 
> > > > Thank you for fixing the historical run!
> > > > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > > > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> > > 
> > > 
> > > Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.
> > 
> > 
> > This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.
> 
> Got it, I see, thanks. Ok, I (a) rolled back the changes I had made to those unit tests, and (b) now if a `baseline_` column was added during the `CombinedDataHandler` estimandization process, it's now deleted. Let me know what you think 😄

Hmm, I am not sure this solved the problem. When I run the testbed with `party_vote_share_dem` as a estimand this is what current results looks like:
```
     geographic_unit_fips    State  ... baseline_party_vote_share_dem  results_party_vote_share_dem
0                   01001  Alabama  ...                      0.239569                      0.239569
1                   01003  Alabama  ...                      0.194804                      0.194804
2                   01005  Alabama  ...                      0.466603                      0.466603
3                   01007  Alabama  ...                      0.214220                      0.214220
4                   01009  Alabama  ...                      0.084699                      0.084699
```

dmnapolitano:
> > > > > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > > > > Also, with a historical election I am getting this error:
> > > > > > > ```
> > > > > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > >     
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >     
> > > > > > >   
> > > > > > > here is an example
> > > > > > > ```
> > > > > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > > > > ```
> > > > > > 
> > > > > > 
> > > > > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > > > > As for your other questions:
> > > > > > 
> > > > > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > > > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > > > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > > > > 
> > > > > 
> > > > > Thank you for fixing the historical run!
> > > > > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > > > > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> > > > 
> > > > 
> > > > Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.
> > > 
> > > 
> > > This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.
> > 
> > 
> > Got it, I see, thanks. Ok, I (a) rolled back the changes I had made to those unit tests, and (b) now if a `baseline_` column was added during the `CombinedDataHandler` estimandization process, it's now deleted. Let me know what you think 😄
> 
> Hmm, I am not sure this solved the problem. When I run the testbed with `party_vote_share_dem` as a estimand this is what current results looks like:
> 
> ```
>      geographic_unit_fips    State  ... baseline_party_vote_share_dem  results_party_vote_share_dem
> 0                   01001  Alabama  ...                      0.239569                      0.239569
> 1                   01003  Alabama  ...                      0.194804                      0.194804
> 2                   01005  Alabama  ...                      0.466603                      0.466603
> 3                   01007  Alabama  ...                      0.214220                      0.214220
> 4                   01009  Alabama  ...                      0.084699                      0.084699
> ```

Oh interesting!  How did you produce this data frame?  (Code, commands, etc.) 🤔

lennybronner:
> > > > > > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > > > > > Also, with a historical election I am getting this error:
> > > > > > > > ```
> > > > > > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > > > > > ```
> > > > > > > > 
> > > > > > > > 
> > > > > > > >     
> > > > > > > >       
> > > > > > > >     
> > > > > > > > 
> > > > > > > >       
> > > > > > > >     
> > > > > > > > 
> > > > > > > >     
> > > > > > > >   
> > > > > > > > here is an example
> > > > > > > > ```
> > > > > > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > > > > > As for your other questions:
> > > > > > > 
> > > > > > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > > > > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > > > > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > > > > > 
> > > > > > 
> > > > > > Thank you for fixing the historical run!
> > > > > > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > > > > > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> > > > > 
> > > > > 
> > > > > Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.
> > > > 
> > > > 
> > > > This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.
> > > 
> > > 
> > > Got it, I see, thanks. Ok, I (a) rolled back the changes I had made to those unit tests, and (b) now if a `baseline_` column was added during the `CombinedDataHandler` estimandization process, it's now deleted. Let me know what you think 😄
> > 
> > 
> > Hmm, I am not sure this solved the problem. When I run the testbed with `party_vote_share_dem` as a estimand this is what current results looks like:
> > ```
> >      geographic_unit_fips    State  ... baseline_party_vote_share_dem  results_party_vote_share_dem
> > 0                   01001  Alabama  ...                      0.239569                      0.239569
> > 1                   01003  Alabama  ...                      0.194804                      0.194804
> > 2                   01005  Alabama  ...                      0.466603                      0.466603
> > 3                   01007  Alabama  ...                      0.214220                      0.214220
> > 4                   01009  Alabama  ...                      0.084699                      0.084699
> > ```
> 
> Oh interesting! How did you produce this data frame? (Code, commands, etc.) 🤔

I ran the model from the testbed (with `estimands = ["dem", "party_vote_share_dem"]`) and put a breakpoint in the CombinedDataHandler to inspect `current_data`

dmnapolitano:
> > > > > > > > > This looks great! I'm curious whether this would work from the testbed? If I understand correctly, then the results estimand is only added from the LiveDataHandler, right? We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?
> > > > > > > > > Also, with a historical election I am getting this error:
> > > > > > > > > ```
> > > > > > > > > KeyError: "['results_party_vote_share_dem'] not in index"
> > > > > > > > > ```
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > >     
> > > > > > > > >       
> > > > > > > > >     
> > > > > > > > > 
> > > > > > > > >       
> > > > > > > > >     
> > > > > > > > > 
> > > > > > > > >     
> > > > > > > > >   
> > > > > > > > > here is an example
> > > > > > > > > ```
> > > > > > > > >  elexmodel 2021-11-02_VA_G --estimands party_vote_share_dem --office_id=G --geographic_unit_type=county --percent_reporting 50 --historical
> > > > > > > > > ```
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Thanks! Ok, so I just pushed changes that should get the `--historical` to work properly. Sorry about that and thanks for providing that example 😅 🎉
> > > > > > > > As for your other questions:
> > > > > > > > 
> > > > > > > > 1. Yes, this should work from the testbed 🤞🏻   However, I'm trying to determine what we should do with unexpected units and a "new_estimand".  How should we populate `results_new_estimand` for unexpected units?  We still want to generate predictions for "new_estimand" for unexpected units, right?
> > > > > > > > 2. The new estimands' `results_` columns should be added from both `LiveDataHandler` and `PreprocessedDataHandler` objects (assuming I did it right 😅).
> > > > > > > > 3. "We might want to add another function that does the same as add_estimand_baselines but just for the results so that it can be called in isolation?"  Sorry, I'm not sure I understand this.  What's a use-case for this? 🤔
> > > > > > > 
> > > > > > > 
> > > > > > > Thank you for fixing the historical run!
> > > > > > > It's not working from the testbed for me though. I think the issue is that the new estimand is never added to the current results (outside of the MockLiveDataHandler -- but that handler is only used together with the cli to test).
> > > > > > > re: unexpected units. Yes we will want to add the new estimand to unexpected units also. We don't make predictions for unexpected units, but we do include their current results in the sums
> > > > > > 
> > > > > > 
> > > > > > Ok, I think I got this working now with the testbed without needing to make any changes to the testbed 🎉 Please try it out and let me know what you think. As a result of the changes I made, I think there are 1-2 additional columns in the `CombinedDataHandler.data` data frames depending on the number of estimands, but nothing else has changed with the unit tests, so I think it's alright 🤔 Open to suggestions, of course.
> > > > > 
> > > > > 
> > > > > This works now. Adding columns to the data is totally fine (we really do want to add a new results column), but I don't think we want to add a new baseline column -- since when merging the preprocessed data and the current we now run into a name collision.
> > > > 
> > > > 
> > > > Got it, I see, thanks. Ok, I (a) rolled back the changes I had made to those unit tests, and (b) now if a `baseline_` column was added during the `CombinedDataHandler` estimandization process, it's now deleted. Let me know what you think 😄
> > > 
> > > 
> > > Hmm, I am not sure this solved the problem. When I run the testbed with `party_vote_share_dem` as a estimand this is what current results looks like:
> > > ```
> > >      geographic_unit_fips    State  ... baseline_party_vote_share_dem  results_party_vote_share_dem
> > > 0                   01001  Alabama  ...                      0.239569                      0.239569
> > > 1                   01003  Alabama  ...                      0.194804                      0.194804
> > > 2                   01005  Alabama  ...                      0.466603                      0.466603
> > > 3                   01007  Alabama  ...                      0.214220                      0.214220
> > > 4                   01009  Alabama  ...                      0.084699                      0.084699
> > > ```
> > 
> > 
> > Oh interesting! How did you produce this data frame? (Code, commands, etc.) 🤔
> 
> I ran the model from the testbed (with `estimands = ["dem", "party_vote_share_dem"]`) and put a breakpoint in the CombinedDataHandler to inspect `current_data`

Ok thanks, I think I got it.  Either I misunderstood what Risha was doing, or she was only creating a column `baseline_foo` for new estimand "foo", no matter what.  So now I (believe I) adjusted the logic to create `results_foo` or `baseline_foo` as necessary.  Let me know what you think 😄 🤞🏻

lennybronner (approved):
LGTM!

PR #61, opened 2023-06-30
ELEX-2402-random-seed-model

rishasurana:
## Description
**Original error description:** In ‘test_redo.py’ and ‘test_vary_params.py’ the unit tests that use the ‘frac_within_pi’ and ‘mean_pi_length’ are currently commented out. That is because those tests rely on some of the randomness in the model (computes CI’s) so won’t be the same value run to run. We need to set a model seed that only affects the unit tests - so those values are the same each time.

**New error description:** The error appears to not be caused by the randomness of the model. This is because the error only occurs during the ‘mean_pi_length’ calculation when the result (on either side, rounded to the second decimal) ends in a 5. The python `.round `function uses "banker's rounding" which is causing the following issues.

ERROR: assert 0.3 == 0.29 <--> 0.296 == 0.295
ERROR: assert 0.26 == 0.25 <--> assert 0.255 == 0.256
ERROR: assert 0.557 == 0.558 <--> 0.56 == 0.558

To resolve this error, a corresponding PR is opened in [elex-live-model-testbed](https://github.com/WPMedia/elex-live-model-testbed/pull/15) to update the tests. This PR will update math_utils.py.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2402

### DIFF
--- a/src/elexmodel/cli.py
+++ b/src/elexmodel/cli.py
@@ -3,14 +3,13 @@ import json
 import click
 from dotenv import find_dotenv, load_dotenv
 
+load_dotenv(find_dotenv())
+
 from elexmodel.client import HistoricalModelClient, ModelClient  # noqa: E402
 from elexmodel.handlers import s3  # noqa: E402
 from elexmodel.handlers.data.LiveData import MockLiveDataHandler  # noqa: E402
 from elexmodel.utils.constants import VALID_AGGREGATES_MAPPING  # noqa: E402
 from elexmodel.utils.file_utils import TARGET_BUCKET  # noqa: E402
-
-load_dotenv(find_dotenv())
-
 
 @click.command()
 @click.argument("election_id")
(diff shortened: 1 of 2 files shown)


lennybronner:
This is kind of incredible, I really appreciate that you were able to figure it out. I'm curious if there are changes we can make in the testbed only that would solve this issue? Since I don't love changing how we are rounding for just computing two kinds of errors.

rishasurana:
> This is kind of incredible, I really appreciate that you were able to figure it out. I'm curious if there are changes we can make in the testbed only that would solve this issue? Since I don't love changing how we are rounding for just computing two kinds of errors.
> 
> Also, why did this error not occur for MAPE unit tests?

It looks like we used `.approx` for the MAPE unit tests- planning to use that for the tests in the testbed to resolve this issue

PR #64, opened 2023-08-04
ELEX-2758-generalize-param-input

rishasurana:
## Description
We deal with all model parameters that get passed into get_estimates the same way, independent of what model they are actually for (non-parameter, gaussian). Instead we should pass a dictionary of additional model parameters (or model settings?) to get_estimates that specify the parameters that are specific to the model pi_method. 

These are:
beta : gaussian
winsorize : gaussian
robust : nonparametric
lambda_ should also be passed in that way (though can be run with either model)

## Jira Ticket
[https://arcpublishing.atlassian.net/browse/ELEX-2758](https://arcpublishing.atlassian.net/browse/ELEX-2758
)
## Test Steps
tox

## AC

1. get_estimates should receive a dictionary with model parameters (ie. the model parameters above)
2. Update _check_input_parameters to make sure that the parameters are included in the specific model. 
3. Update election calls (in unit tests, etc.)

## Notes
Added [C901](https://www.flake8rules.com/rules/C901.html) after AC #2. The model_parameters checks can be moved to another function if we want to keep this flag.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,4 +36,4 @@ repos:
           # these are errors that will be ignored by flake8
           # definitions here
           # https://flake8.pycqa.org/en/latest/user/error-codes.html
-          - "--ignore=E266,E501,W503,C901"
+          - "--ignore=E266,E501,W503,F811,C901"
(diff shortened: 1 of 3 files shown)


jchaskell (approved):
lgtm

PR #70, opened 2023-08-31
Added generalized model parameters to cli

lennybronner:
## Description
[Previously](https://github.com/washingtonpost/elex-live-model/pull/64) we had changed how we pass model specific parameters. Instead of having to define and deal with each of the model specific parameters separately, we now use a dictionary that can have any key, value pair for a model parameter and it's value. This means that if we add a parameter in the future we don't need to create variables for it in the client but it can be passed directly to the model using `model_settings`.

We had forgotten to add this functionality to the cli, which I do with this PR. I also updated the model client a bit to take full advantage of this change and updated the `GaussianModel` for that too.

## Test Steps
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=20 --model_parameters='{"robust":"True"}'
```
and 
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=20 --model_parameters='{"winsorize":"True", "beta": 1, "lambda": 100}'
```
Unit tests have not changed and should all still pass with `Tox`


dmnapolitano:
This looks good to me! 🎉  I have a branch with a fix to get the `cli` to find my `.env` file which I'll need to be able to test this out.  IMO I say we merge this, then I do some testing, then if there's anything weird, I can correct it in that branch and submit the PR 😄

dmnapolitano (approved):

PR #71, opened 2023-08-31
Onboarding documentation and CLI fixes

dmnapolitano:
## Description

Hi! 👋🏻  The changes in this PR:

* in `cli.py`, check if `find_dotenv()` is blank, then check the current working directory with `usecwd=True`.  Note by default, `usecwd=False` and that this change is only necessary on the command-line.  If you're running from an interpreter (`python` or `ipython`) or Jupyter notebook, there's logic in the `dotenv` code to check for this and then check in the CWD. 🤷🏻‍♀️ 
* clarify in the README that one should not `pip install elex-model` if they're doing dev work, like I did 😅  I also converted ordinal lists to numbers instead of bullets.
* gets the `test-model` notebook to work again after getting out-of-sync with changes to `CombinedDataHandler` and the `geographic_unit_type` argument.  Also in this notebook, I reduced some of the duplicate code and added `%config InlineBackend.figure_format = "retina"` which is an option to make graphs look moderately better on PC monitors and SIGNIFICANTLY better on Mac monitors 🤩 
* modify `tox` to run tests with both Python 3.10 and 3.11.

## Jira Ticket

Part of ELEX-3101, Diane's onboarding.

## Test Steps

`tox`, run the `test-model` notebook, and/or run your favorite `elexmodel` command.

## Question

I noticed a lot of the `elexmodel` requirements are pretty old, for example `pandas<1.5` and they're currently up to 2.1 😅  Is there any reason or would you like me to try to update the packages (in another PR of course)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.out
 *.egg-info
 *.eggs
+*~
 
 build/
 dist/
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ The first iteration of this model is written in R in [this repo](https://github.
 
 ## How to Run the Model
 
+For dev work, skip this section and go on to the [Development](##Development) section below.
+
 ### Installation
 
 * We recommend that you set up a virtualenv and activate it (IE ``mkvirtualenv elex-model`` via http://virtualenvwrapper.readthedocs.io/en/latest/).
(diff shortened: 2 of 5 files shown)


lennybronner (approved):
This looks great, and ran just fine!

Upgrading the packages would be a great idea! (and might help us with the warnings in the unit tests)

PR #72, opened 2023-09-01
Elex 2766 separate out conformalization models

lennybronner:
## Description
In order to prepare for the Bootstrap model (which does not use conformalization), I had to change the inheritance structure of our models so far. We now have a `ConformalElectionModel` which inherits from the `BaseElectionModel` and the `GaussianElectionModel` and `NonparametricElectionModel` inherit from that. This means we can re-use the aggregation code from the `BaseElectionModel` in the upcoming bootstrap model without running into issues with a non-existent conformalization set. 

We now only deal with the conformalization data in the client if we are using a `ConformalElectionModel` rather than always, which is what we were doing previously.

I also added type hints and updated the unit tests correspondingly.

## Jira Ticket
[elex-2766 ](https://arcpublishing.atlassian.net/browse/ELEX-2766)

## Test Steps
Unit tests are updated, so `tox` works.

Also both old models work like normal:
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=gaussian --percent_reporting=20 --model_parameters="{'lambda_': 1, 'beta': 4, 'winsorize': True}" --aggregates=postal_code --aggregates=county_classification --aggregates=unit
```
and 
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=20 --model_parameters="{'lambda_': 1, 'robust': True}"
```

### DIFF
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -141,7 +141,7 @@ class BaseElectionModel:
         """
         pass
 
-    def get_coefficients(self):
+    def get_coefficients(self) -> dict:
         """
         Returns a dictionary of feature/fixed effect names to the coefficients
         These coefficients are for the point prediciton only.
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -18,6 +18,12 @@ class BaseElectionModel:
         self.add_intercept = True
         self.seed = model_settings.get("seed", 4191)
 
+    def get_minimum_reporting_units(self, alpha: float) -> int:
+        """
+        Returns the minimum number of units necessary to run the model
+        """
+        return 10
+
     def get_unit_predictions(self, reporting_units: pd.DataFrame, nonreporting_units: pd.DataFrame, estimand: str, *kwargs) -> np.ndarray:
         """
         Generates and returns unit level predictions
--- a/src/elexmodel/models/BaseElectionModel.py
+++ b/src/elexmodel/models/BaseElectionModel.py
@@ -1,94 +1,31 @@
 import logging
-import math
-import warnings
 from collections import namedtuple
 
-import cvxpy
+import pandas as pd
 import numpy as np
-from elexsolver.QuantileRegressionSolver import QuantileRegressionSolver
-
-from elexmodel.handlers.data.Featurizer import Featurizer
-
-warnings.filterwarnings("error", category=UserWarning, module="cvxpy")
-
-PredictionIntervals = namedtuple("PredictionIntervals", ["lower", "upper", "conformalization"], defaults=(None,) * 3)
 
 LOG = logging.getLogger(__name__)
 
+PredictionIntervals = namedtuple("PredictionIntervals", ["lower", "upper"], defaults=(None,) * 2)
+
 
 class BaseElectionModel:
-    def __init__(self, model_settings={}):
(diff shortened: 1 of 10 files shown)


src/elexmodel/models/BaseElectionModel.py
```diff
@@ -1,94 +1,40 @@
...
-
-        # add in last election results to go from residual to number of votes in this election
-        # max with results so that predictions are always at least ars large as actual number of votes
-        preds = np.maximum(
-            preds + nonreporting_units[f"last_election_results_{estimand}"], nonreporting_units[f"results_{estimand}"]
-        )
-
-        # round since we don't need the artificial precision
-        return preds.round(decimals=0)
+        pass
```
dmnapolitano, line 33: Is the idea that the `BaseElectionModel` can't be instantiated and one has to use a subclass instead?  Because in that case, I'd recommend creating it as an abstract base class and have this (and similar) method(s) return `NotImplemented` so that an exception is raised: https://docs.python.org/3/library/abc.html 🤔

src/elexmodel/models/ConformalElectionModel.py
```diff
@@ -0,0 +1,194 @@
...
+        Returns conformalization data at the unit level
+        Conformalization data is adjustment from estimated % change from baseline
+        """
+        pass
+
+    def get_all_conformalization_data_agg(self):
+        """
+        Returns conformalization data at the aggregate level
+        """
+        pass
```
dmnapolitano, line 187: Why are these two methods just pass-throughs? 🤔

src/elexmodel/models/BaseElectionModel.py
```diff
@@ -1,94 +1,40 @@
...
-    def fit_model(self, model, df_X, df_y, tau, weights, normalize_weights):
+    def get_minimum_reporting_units(self, alpha: float) -> int:
         """
-        Fits the quantile regression for the model
+        Returns the minimum number of units necessary to run the model
         """
-        X = df_X.values
-        y = df_y.values
-        weights = weights.values
+        return 10
```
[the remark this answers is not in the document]
lennybronner: In effect the number has to be greater than the number of features + fixed effects (with some slight modification in the case of conformalization). So in that case 10 is entirely arbitrary. I will change it to NotImplemented like your suggestion below!

src/elexmodel/models/BaseElectionModel.py
lennybronner: yes, that is the idea! That's a great idea, will do -- thanks!

src/elexmodel/models/ConformalElectionModel.py
lennybronner: same as above. will change to NotImplemented

src/elexmodel/models/BaseElectionModel.py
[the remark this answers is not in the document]
lennybronner: I think that would be ideal, but we don't have access to the final design matrix in this function so we don't actually know the number of columns 🙃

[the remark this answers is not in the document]
dmnapolitano: Oh lol 😭

lennybronner:
> A couple of questions and suggestions but otherwise, I think this looks great! 🎉

Ok, I updated the two new models into abstract base classes and added/updated/moved unit tests correspondingly!

PR #77, opened 2023-09-22
Elex 2771 new ols qr

lennybronner:
## Description
This PR moves over parts of the changes we are making for the [bootstrap election model PR](https://github.com/washingtonpost/elex-live-model/pull/76) in order to make reviewing that one easier. It make the changes necessary to the old `ConformalElectionModel` to work with the updates made to elex-solver [in this PR](https://github.com/washingtonpost/elex-solver/pull/11) and it makes small tweaks to the estimandizer to prepare it for multiple estimands being generated at once. It also updates unit tests accordingly.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-2771

## Test Steps
`tox`

also 
```
elexmodel 2020-11-03_USA_G --office_id=P --estimands=dem --geographic_unit_type=county --pi_method=nonparametric --percent_reporting=30 --aggregates=postal_code --historical
```

### DIFF
--- a/src/elexmodel/handlers/data/Estimandizer.py
+++ b/src/elexmodel/handlers/data/Estimandizer.py
@@ -1,4 +1,4 @@
-from numpy import nan
+import numpy as np
 
 RESULTS_PREFIX = "results_"
 BASELINE_PREFIX = "baseline_"
@@ -11,12 +11,15 @@ class Estimandizer:
 
     def add_estimand_results(self, data_df, estimands, historical):
         columns_to_return = []
+        turnout_col = f"{RESULTS_PREFIX}turnout"
+
         for estimand in estimands:
             results_col = f"{RESULTS_PREFIX}{estimand}"
+            additional_columns_added = []
             if results_col not in data_df.columns:
                 # will raise a KeyError if a function with the same name as `estimand` doesn't exist
                 try:
-                    data_df = globals()[estimand](data_df, RESULTS_PREFIX)
+                    data_df, additional_columns_added = globals()[estimand](data_df, RESULTS_PREFIX)
                 except KeyError as e:
                     if historical:
                         # A historical run is one where we pull in data from a past election
@@ -27,12 +30,21 @@ class Estimandizer:
                         # and that data handler expects a results_ column for every estimand specified.
                         # Hence, this is the only special case in which we'd want to add
                         # an empty results_ column.
-                        data_df[results_col] = nan
+                        data_df[results_col] = np.nan
                     else:
                         # If this is not a historical run, then this is a live election
                         # so we are expecting that there will be actual results data
                         raise e
-            columns_to_return.append(results_col)
+
+
+            columns_to_return.extend([results_col] + additional_columns_added)
+
+        # always adding turnout since we will want to generate weights
+        # but if turnout is the estimand, then we only want to add it once
+        if turnout_col not in columns_to_return:
+            columns_to_return.append(turnout_col)
+        
+        data_df = self.add_weights(data_df, RESULTS_PREFIX)
 
         return data_df, columns_to_return
 
@@ -49,7 +61,7 @@ class Estimandizer:
             baseline_col = f"{BASELINE_PREFIX}{pointer}"
 
             if baseline_col not in data_df.columns:
-                data_df = globals()[estimand](data_df, BASELINE_PREFIX)
+                data_df, __ = globals()[estimand](data_df, BASELINE_PREFIX)
 
             if not historical:
                 data_df[f"last_election_results_{estimand}"] = data_df[baseline_col].copy() + 1
@@ -62,6 +74,15 @@ class Estimandizer:
             # we need to add the results from the historical election as well.
             data_df, ___ = self.add_estimand_results(data_df, estimand_baselines.keys(), historical)
 
+        data_df = self.add_weights(data_df, BASELINE_PREFIX)
+        return data_df
+
+    def add_weights(self, data_df, col_prefix):
+        data_df[f"{col_prefix}weights"] = data_df[f"{col_prefix}turnout"]
+        return data_df
+
+    def add_turnout_factor(self, data_df):
+        data_df["turnout_factor"] = np.nan_to_num(data_df.results_weights / data_df.baseline_weights)
         return data_df
 
 
@@ -76,4 +97,4 @@ def party_vote_share_dem(data_df, col_prefix):
         lambda x: 0 if x[numer] == 0 or x[denom] == 0 else x[numer] / x[denom], axis=1
     )
 
-    return data_df
+    return data_df, []
--- a/src/elexmodel/handlers/data/Estimandizer.py
+++ b/src/elexmodel/handlers/data/Estimandizer.py
@@ -52,7 +52,7 @@ class Estimandizer:
         # if we are in a historical election we are only reading preprocessed data to get
         # the historical election results of the currently reporting units.
         # so we don't care about the total voters or the baseline election.
-        
+
         data_df = self.add_weights(data_df, BASELINE_PREFIX)
 
         for estimand, pointer in estimand_baselines.items():
--- a/src/elexmodel/handlers/data/Estimandizer.py
+++ b/src/elexmodel/handlers/data/Estimandizer.py
@@ -97,5 +97,7 @@ class Estimandizer:
 
 
 def party_vote_share_dem(data_df, col_prefix):
-    data_df[f"{col_prefix}party_vote_share_dem"] = np.nan_to_num(data_df[f"{col_prefix}dem"] /data_df[f"{col_prefix}turnout"])
+    data_df[f"{col_prefix}party_vote_share_dem"] = np.nan_to_num(
+        data_df[f"{col_prefix}dem"] / data_df[f"{col_prefix}turnout"]
+    )
     return data_df, []
--- a/src/elexmodel/handlers/data/Estimandizer.py
+++ b/src/elexmodel/handlers/data/Estimandizer.py
@@ -13,6 +13,8 @@ class Estimandizer:
         columns_to_return = []
         turnout_col = f"{RESULTS_PREFIX}turnout"
 
+        data_df = self.add_weights(data_df, RESULTS_PREFIX)
+
         for estimand in estimands:
             results_col = f"{RESULTS_PREFIX}{estimand}"
             additional_columns_added = []
@@ -31,6 +33,7 @@ class Estimandizer:
                         # Hence, this is the only special case in which we'd want to add
                         # an empty results_ column.
                         data_df[results_col] = np.nan
+                        data_df[turnout_col] = np.nan
                     else:
                         # If this is not a historical run, then this is a live election
                         # so we are expecting that there will be actual results data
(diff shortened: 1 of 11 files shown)


dmnapolitano:
@lennybronner thank you so much for all of this! 🎉 🙌🏻 

Can you say a bit about what the turnout factor is? 😅

dmnapolitano:
> @lennybronner thank you so much for all of this! 🎉 🙌🏻
> 
> Can you say a bit about what the turnout factor is? 😅

Or actually, sorry, I think I see what you're doing.  Even though the end result we're sharing with the world is the predicted `margin`, we still need to predict `turnout` in order to predict `margin`.  But not all of our data sets include `results_turnout`, so you're checking to make sure it exists and if not, sum across the (applicable) `results_` columns we do have.  Do I have that right-ish? 😅

lennybronner:
> > @lennybronner thank you so much for all of this! 🎉 🙌🏻
> > Can you say a bit about what the turnout factor is? 😅
> 
> Or actually, sorry, I think I see what you're doing. Even though the end result we're sharing with the world is the predicted `margin`, we still need to predict `turnout` in order to predict `margin`. But not all of our data sets include `results_turnout`, so you're checking to make sure it exists and if not, sum across the (applicable) `results_` columns we do have. Do I have that right-ish? 😅

Yeah, we need to predict turnout in order to get the normalization constant for normalized margin, since we need to go back and forth between unnormalized and normalized margin to move from county predictions to state predictions.

Turnout factor is basically just the ratio of turnout in this election to turnout in last election. In the margin model it's part of what we're estimating. But we also drop units whose turnout factor is greater than or less than some constant. We're basically assuming that if turnout in some county is only 20% of it's last elections turnout (or greater than 200% of last election's turnout) that our results provider either made a mistake (or that we accidentally mismatched precincts), so we drop that county in our model. We can adjust the constants (20/200%) through parameters in the model so in case that there is a super low/high turnout election we don't accidentally drop too many units.

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -11,12 +11,15 @@ class Estimandizer:
...
+        turnout_col = f"{RESULTS_PREFIX}turnout"
+
         for estimand in estimands:
             results_col = f"{RESULTS_PREFIX}{estimand}"
+            additional_columns_added = []
             if results_col not in data_df.columns:
                 # will raise a KeyError if a function with the same name as `estimand` doesn't exist
                 try:
-                    data_df = globals()[estimand](data_df, RESULTS_PREFIX)
+                    data_df, additional_columns_added = globals()[estimand](data_df, RESULTS_PREFIX)
```
dmnapolitano, line 22: Sorry, I'm not sure I understand why some estimandized estimands return `additional_columns_added` and some don't 🤔

src/elexmodel/handlers/data/Estimandizer.py
```diff
@@ -62,6 +74,22 @@ def add_estimand_baselines(self, data_df, estimand_baselines, historical, includ
...
+    def add_weights(self, data_df, col_prefix):
+        data_df[f"{col_prefix}weights"] = data_df[f"{col_prefix}turnout"]
+        return data_df
+
+    def add_turnout_factor(self, data_df):
+        # posinf and neginf are also set to zero because dividing by zero can lead to nan/posinf/neginf depending
+        # on the type of the numeric in the numpy array. Assume that if baseline_weights is zero then turnout
+        # would be incredibly low in this election too (ie. this is effectively an empty precinct) and so setting
+        # the turnout factor to zero is fine
+        data_df["turnout_factor"] = np.nan_to_num(
```
dmnapolitano, line 90: Oh wow, I didn't know this existed!  This is exactly what I was trying to do in `party_vote_share_dem` with the `.apply()` 😅 🎉

src/elexmodel/models/ConformalElectionModel.py
```diff
@@ -185,8 +186,8 @@ def get_unit_prediction_interval_bounds(
         # are the same accross train_data, conformalization_data and nonreporting_units
         nonreporting_units_features = interval_featurizer.generate_holdout_data(x_all[self.n_train :])  # noqa: E203
 
-        nonreporting_lower_bounds = lower_qr.predict(nonreporting_units_features)
-        nonreporting_upper_bounds = upper_qr.predict(nonreporting_units_features)
+        nonreporting_lower_bounds = lower_qr.predict(nonreporting_units_features.values).flatten()
+        nonreporting_upper_bounds = upper_qr.predict(nonreporting_units_features.values).flatten()
```
dmnapolitano, line 190: Interesting, I didn't realize there was a change in the predicted output data structure that would necessitate `.flatten()`s here (and elsewhere) 🤔

tests/handlers/test_combined_data.py
```diff
@@ -7,58 +7,104 @@
...
+        office_id,
+        geographic_unit_type,
+        estimand_baselines=estimand_baselines,
+        estimands=estimands,
+        data=va_governor_county_data,
+    )
+    preprocessed_data = preprocessed_data_handler.data
 
     combined_data_handler = CombinedDataHandler(
-        va_governor_county_data, current_data, estimands, "county", handle_unreporting="drop"
```
dmnapolitano, line 17: Is there a reason why we can't reuse or update `va_governor_county_data` to be the required `PreprocessedData` here (and in other similar unit tests)?  I think this test is testing too many things, as in, `PreprocessedDataHandler` and `CombinedDataHandler` when it should only be testing `CombinedDataHandler`.  Or we could create another test 🤔

tests/handlers/test_combined_data.py
```diff
@@ -7,58 +7,104 @@
...
+        preprocessed_data, current_data, estimands, "county", handle_unreporting="drop"
     )
-    assert combined_data_handler.data.shape == (133, 29)
+
+    assert combined_data_handler.data.shape == (133, 33)
```
dmnapolitano, line 33: I'm seeing increases of 4-6 columns for a lot of unit tests like this one.  Can you explain why?  I'm trying to think through the columns that are being added now.  Possibly `baseline_turnout` and `results_turnout` but that's all I can think of 🤔

dmnapolitano (changes requested):
Just some questions and questions-that-may-become-suggestions lol, otherwise looks good! 😄 🎉

lennybronner:
> > > > @lennybronner thank you so much for all of this! 🎉 🙌🏻
> > > > Can you say a bit about what the turnout factor is? 😅
> > > 
> > > 
> > > Or actually, sorry, I think I see what you're doing. Even though the end result we're sharing with the world is the predicted `margin`, we still need to predict `turnout` in order to predict `margin`. But not all of our data sets include `results_turnout`, so you're checking to make sure it exists and if not, sum across the (applicable) `results_` columns we do have. Do I have that right-ish? 😅
> > 
> > 
> > Yeah, we need to predict turnout in order to get the normalization constant for normalized margin, since we need to go back and forth between unnormalized and normalized margin to move from county predictions to state predictions.
> > Turnout factor is basically just the ratio of turnout in this election to turnout in last election. In the margin model it's part of what we're estimating. But we also drop units whose turnout factor is greater than or less than some constant. We're basically assuming that if turnout in some county is only 20% of it's last elections turnout (or greater than 200% of last election's turnout) that our results provider either made a mistake (or that we accidentally mismatched precincts), so we drop that county in our model. We can adjust the constants (20/200%) through parameters in the model so in case that there is a super low/high turnout election we don't accidentally drop too many units.
> 
> Got it!! That's awesome 🎉
> 
> What about dropping units whose turnout factors are outliers against the other units? That way, on the off chance the entire state doesn't vote (or does vote), there's no risk of dropping almost every unit in the state. If you've done some evaluation to come up with these constants, that's fine, and I know for now we're primarily interested in big (top-of-the-) ticket races anyway where this is less likely to occur. Just a thought 🤷🏻‍♀️ 😄

That's a really good idea! Though I guess would necessitate a bit more computation? Do you mind adding a future ticket to implement?

src/elexmodel/handlers/data/Estimandizer.py
lennybronner: Ah, this is in preparation for margin. We'll be adding both normalized and un-normalized margin to the dataframe. So this just gives us the option to do it.

src/elexmodel/handlers/data/Estimandizer.py
lennybronner: I pushed an update to that function to use this if that's ok?

src/elexmodel/models/ConformalElectionModel.py
lennybronner: yes! This is because of the changes we made in elex solver for the quantile regression. We now have this for-loop where quantile regression iterates over the different quantiles and adds the coefficients to a list. That new outer list is what adds the extra dimension that we want to get rid of here.

tests/handlers/test_combined_data.py
lennybronner: good call! I updated the tests so now we modify `va_governor_county_data` so that it has the columns that `CombinedDataHandler` expects

tests/handlers/test_combined_data.py
lennybronner: yes! it's turnout factor and results turnout and the two weights columns (results and baseline)

dmnapolitano:
> > > > > @lennybronner thank you so much for all of this! 🎉 🙌🏻
> > > > > Can you say a bit about what the turnout factor is? 😅
> > > > 
> > > > 
> > > > Or actually, sorry, I think I see what you're doing. Even though the end result we're sharing with the world is the predicted `margin`, we still need to predict `turnout` in order to predict `margin`. But not all of our data sets include `results_turnout`, so you're checking to make sure it exists and if not, sum across the (applicable) `results_` columns we do have. Do I have that right-ish? 😅
> > > 
> > > 
> > > Yeah, we need to predict turnout in order to get the normalization constant for normalized margin, since we need to go back and forth between unnormalized and normalized margin to move from county predictions to state predictions.
> > > Turnout factor is basically just the ratio of turnout in this election to turnout in last election. In the margin model it's part of what we're estimating. But we also drop units whose turnout factor is greater than or less than some constant. We're basically assuming that if turnout in some county is only 20% of it's last elections turnout (or greater than 200% of last election's turnout) that our results provider either made a mistake (or that we accidentally mismatched precincts), so we drop that county in our model. We can adjust the constants (20/200%) through parameters in the model so in case that there is a super low/high turnout election we don't accidentally drop too many units.
> > 
> > 
> > Got it!! That's awesome 🎉
> > What about dropping units whose turnout factors are outliers against the other units? That way, on the off chance the entire state doesn't vote (or does vote), there's no risk of dropping almost every unit in the state. If you've done some evaluation to come up with these constants, that's fine, and I know for now we're primarily interested in big (top-of-the-) ticket races anyway where this is less likely to occur. Just a thought 🤷🏻‍♀️ 😄
> 
> That's a really good idea! Though I guess would necessitate a bit more computation? Do you mind adding a future ticket to implement?

Sure!  Thanks! 😄 🎉  The ticket is here: https://arcpublishing.atlassian.net/browse/ELEX-3298

tests/handlers/test_combined_data.py
dmnapolitano: Nice!  Thanks! 🎉 🎉

dmnapolitano (approved):
I think this looks good! 🎉 🎉

PR #86, opened 2023-11-01
some model parameters now also allow ints and not just floats

lennybronner:
## Description
Some model parameters should be numeric but are currently only accepted as floats by the model. This is causing issues with hasura, which converts floats to ints if they have a trailing zero if you change the parameters directly in the database. So this updates the model to let the parameters be both floats and ints.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-3460

## Test Steps
The following is throwing an error in develop branch, but not in this branch:
```
elexmodel 2017-11-07_VA_G --estimands=margin --office_id=G --geographic_unit_type=precinct --pi_method=bootstrap --percent_reporting=10 --aggregates=postal_code --fixed_effects=county_classification --features=baseline_normalized_margin  --model_parameters='{"B": 10, "y_unobserved_upper_bound": 2}'
```

### DIFF
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -104,11 +104,11 @@ class ModelClient:
             ):
                 raise ValueError("lambda is not valid. It has to be numeric and greater than zero.")
             if "turnout_factor_lower" in model_parameters and not isinstance(
-                model_parameters["turnout_factor_lower"], float
+                model_parameters["turnout_factor_lower"], (float, int)
             ):
                 raise ValueError("turnout_factor_lower is not valid. Has to be a float.")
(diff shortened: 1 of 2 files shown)


dmnapolitano (commented):
This looks good!  Thanks for figuring this out! 🎉 

Should we add a unit test? 🤔

lennybronner:
> This looks good! Thanks for figuring this out! 🎉
> 
> Should we add a unit test? 🤔

Added unit tests!

dmnapolitano (approved):
LGTM!  Thanks for adding the unit test! 🎉 🎉

PR #87, opened 2023-11-02
fixed alpha bug

lennybronner:
## Description
The bootstrap model function `get_national_summary_estimates` expected a parameter called `alpha` that set the size of the prediction interval. But the model client function `get_national_summary_votes_estimates` that invokes the model function didn't pass in an alpha.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-3461

## Test Steps
In the testbed run the model with `agg_model_preds=True` in develop, which should break. It should work in this branch.

### DIFF
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -169,8 +169,8 @@ class ModelClient:
         raw_aggregate_list = base_aggregate + [aggregate]
         return sorted(list(set(raw_aggregate_list)), key=lambda x: AGGREGATE_ORDER.index(x))
 
-    def get_national_summary_votes_estimates(self, nat_sum_data_dict=None, called_states={}, base_to_add=0):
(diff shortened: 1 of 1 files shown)


dmnapolitano (changes requested):
Good catch!  Sorry I missed this before 😅 🎉 

TBH on the other branch, I don't really think we need a unit test for that and I'd be fine merging it without it.  But for this change, 😬  Although, I just noticed we don't have one for `get_national_summary_votes_estimates()` without `alpha`, so maybe I should just create a Jira ticket to go through `client` and add missing unit tests at a later date.

What do you think? 🤔

dmnapolitano (approved):
This LGTM!! 🎉   Related Jira ticket to add more unit tests and get the aggregate model working via CLI is here: https://arcpublishing.atlassian.net/browse/ELEX-3462 (to be resolved either by splitting this Jira ticket into two or two branches 😅 )

PR #90, opened 2023-11-03
changed dict print

lennybronner:
## Description
We were printing the error dictionary in the called_states length error for the aggregate model


anthonyjpesce (approved):
lgtm

PR #95, opened 2024-08-01
Fix warnings, update packages, get unit tests to pass again

dmnapolitano:
## Description

Hi!  The changes in this PR:

* update our dependencies to their latest versions,
* fix warnings brought about by virtue of updating said dependencies,
* ensure that our unit tests pass again,
* remove Python 3.10 from unit testing.

Note that there is still a warning coming from `cvxpy` regarding the fact that in `elex-solver`, we haven't explicitly specified which solver to use.  That will require a new `elex-solver` release at a later date; see [ELEX-4413].

## Jira Ticket

[ELEX-3581]

## Test Steps

`tox` 🎉 

[ELEX-4413]: https://arcpublishing.atlassian.net/browse/ELEX-4413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ELEX-3581]: https://arcpublishing.atlassian.net/browse/ELEX-3581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/models/test_nonparametric_election_model.py
+++ b/tests/models/test_nonparametric_election_model.py
@@ -206,10 +206,10 @@ def test_aggregation_simple():
     df3 = model._get_reporting_aggregate_votes(df1, df2, aggregate=["c1", "c2"], estimand=estimand)
     assert pd.DataFrame(
         {
-            "c1": ["a", "a", "b", "b", "a", "d"],
-            "c2": ["x", "y", "y", "z", "w", "t"],
-            f"results_{estimand}": [5.0, 9.0, 1.0, 15.0, 5.0, 1.0],
-            "reporting": [2.0, 1.0, 1.0, 3.0, 1.0, 1.0],
+            "c1": ["a", "a", "a", "b", "b", "d"],
+            "c2": ["w", "x", "y", "y", "z", "t"],
+            f"results_{estimand}": [5.0, 5.0, 9.0, 1.0, 15.0, 1.0],
+            "reporting": [1.0, 2.0, 1.0, 1.0, 3.0, 1.0],
         }
     ).equals(df3)
 
--- a/setup.py
+++ b/setup.py
@@ -4,12 +4,12 @@ from codecs import open
 from setuptools import find_packages, setup
 
 INSTALL_REQUIRES = (
-    "click~=8.1",
+    "click>=8.1",
     "elex-solver>=2.0.1",
-    "pandas~=2.1",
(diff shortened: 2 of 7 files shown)


tests/models/test_nonparametric_election_model.py
```diff
@@ -206,10 +206,10 @@ def test_aggregation_simple():
     df3 = model._get_reporting_aggregate_votes(df1, df2, aggregate=["c1", "c2"], estimand=estimand)
     assert pd.DataFrame(
```
lennybronner, line 207: why are the changes to this unit test necessary?

lennybronner (dismissed):

tests/models/test_nonparametric_election_model.py
dmnapolitano: There's no guarantee (starting with the most recent version of `pandas`, I believe) that the order of the rows in both data frames will be preserved during a `merge()` (or `join()`).  By default they're now being sorted by the columns being merged on, and oddly enough I tried to disable that (with `sort=False`) but it made no difference 🤔 

But one thing I've been concerned about here as opposed to the test I had to modify in `test_gaussian_model.py`....with that test I was able to set the index and compare each value index-by-index.  So as a result, the sort order in the merged data frame no longer matters and our unit test should continue to pass.  However, I'm not doing that here, so, since the result of this unit test depends on the order of the merged data frame, 😬 

I could modify this to set `c1` and `c2` to be the index and check each `results_estimand` and `reporting` columns' values that way 🤔   (Another option is https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html, however, it seems like the index has to match regardless of the row or column order 🤔 )

lennybronner: interesting. ok, will approve. but the order not being guarnteed is def something we should remember. It may require changes to the model elsewhere if we are assuming the order (ie alphabetical or by fips code or something)

lennybronner (approved):

tests/models/test_nonparametric_election_model.py
dmnapolitano: Word 😬

PR #99, opened 2024-08-19
Small agg model bug

lennybronner:
## Description
As part of the hunt for the [2023 agg model bug ](https://arcpublishing.atlassian.net/browse/ELEX-3506) we identified a different, smaller bug that effects the intervals only (not the point prediction). On occasion the double sampling can create intervals that are slightly too wide.

We max the agg model predictions with the number of contests won by the LHS party since that is the minimum number of contests that party can win. We also min with the number of contests won by the RHS party since that is the maximum number they can win.

Also small change to finding the uncalled races to avoid floating point issues.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-4453

## Test Steps
```
python run.py 2023-11-07_VA_G redo --office_id Y --geographic_unit_type county-district --pi_method bootstrap --estimands "['margin']" --features "['baseline_normalized_margin']" --start_timestamp "2023-11-07 19:00:00+01:00" --current_run_name va-y-test --percent_ev_threshold 95 --agg_model_preds --base_to_add 20
```
in the testbed

### DIFF
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -1420,7 +1420,9 @@ class BootstrapElectionModel(BaseElectionModel):
         # for contests called for the RHS party. So this sum gives us the number of called contests for RHS
         called_contests_rhs = np.nansum(1 - called_states_sorted_vals)
 
-        # Our values should not be lower than the number of called
+        # Since the values should be greater than the called_contests_lhs we max with that and since they
+        # should be less than the called_contests_rhs we min with that. Also we add  in the base to account
+        # for uncontested races.
         agg_pred = min(max(aggregate_dem_vals_pred, called_contests_lhs), called_contests_rhs) + base_to_add
         agg_lower = min(max(interval_lower, called_contests_lhs), called_contests_rhs) + base_to_add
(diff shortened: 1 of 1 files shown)


src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -1407,12 +1407,32 @@ def get_national_summary_estimates(
...
+            (1 - called_states_sorted_vals) * nat_sum_data_dict_sorted_vals
+        )
+
+        # Since the values should be greater than the called_values_lhs we max with that and since they
+        # should be less than the called_values_rhs we min with that. Also we add  in the base to account
+        # for uncontested races.
+        agg_pred = min(max(aggregate_dem_vals_pred, called_values_lhs), called_values_rhs) + base_to_add
+        agg_lower = min(max(interval_lower, called_values_lhs), called_values_rhs) + base_to_add
+        agg_upper = min(max(interval_upper, called_values_lhs), called_values_rhs) + base_to_add
+        national_summary_estimates = {"margin": [agg_pred, agg_lower, agg_upper]}
```
dmnapolitano, line 1436: Very helpful explanation! 🎉

dmnapolitano (approved):
This LGTM!  Thanks for tracking this problem down 😄 👏🏻

PR #100, opened 2024-09-03
ELEX-3469 save aggregate predictions to s3

dmnapolitano:
## Description

Hi!  The changes in this PR allow us to save aggregate model output (national summary estimates) to s3.  It also provides a new CLI argument, `--national_summary`, which will produce aggregate model output via CLI 🎉   Hopefully I did this right 😬  Thanks!

## Jira Ticket

[ELEX-3469]

## Test Steps

Set `APP_ENV=dev` and `DATA_ENV=dev`, then run any command with `--national_summary`, e.g.:

```
elexmodel 2023-11-07_VA_G --estimands=margin --features=baseline_normalized_margin --office_id=Y --geographic_unit_type=county-district --save_output results --pi_method bootstrap --national_summary
```

[ELEX-3469]: https://arcpublishing.atlassian.net/browse/ELEX-3469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -174,7 +174,6 @@ class ModelClient:
         raw_aggregate_list = base_aggregate + [aggregate]
         return sorted(list(set(raw_aggregate_list)), key=lambda x: AGGREGATE_ORDER.index(x))
 
-
     def get_national_summary_votes_estimates(self, nat_sum_data_dict=None, base_to_add=0, alpha=0.99):
         if self.model is None:
             raise ModelClientException(
@@ -190,7 +189,6 @@ class ModelClient:
             )
 
         return nat_sum_estimates
-
 
     def get_estimates(
         self,
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -42,7 +42,10 @@ class ModelClient:
         self.all_conformalization_data_agg_dict = defaultdict(dict)
         self.model = None
         self.results_handler = None
-        self._model_updated = False
+        self.election_id = None
+        self.office = None
+        self.geographic_unit_type = None
+        self.save_results = None
 
     def _check_input_parameters(
         self,
@@ -172,32 +175,26 @@ class ModelClient:
         return sorted(list(set(raw_aggregate_list)), key=lambda x: AGGREGATE_ORDER.index(x))
 
     def get_national_summary_votes_estimates(
-        self, nat_sum_data_dict=None, called_states=None, base_to_add=0, alpha=0.99
+        self,
+        nat_sum_data_dict=None,
+        called_states=None,
+        base_to_add=0,
+        alpha=0.99,
     ):
         if self.model is None:
             raise ModelClientException(
                 "Must call the get_estimands() method before get_national_summary_votes_estimates()."
             )
 
-        if self._model_updated or "nat_sum_data" not in self.results_handler.final_results:
-            nat_sum_estimates = self.model.get_national_summary_estimates(
-                nat_sum_data_dict, called_states, base_to_add, alpha
-            )
-            self.results_handler.add_national_summary_estimates(nat_sum_estimates)
-            self._model_updated = False
+        nat_sum_estimates = self.model.get_national_summary_estimates(
+            nat_sum_data_dict, called_states, base_to_add, alpha
+        )
+        self.results_handler.add_national_summary_estimates(nat_sum_estimates)
 
-        else:
-            nat_sum_estimates = (
-                self.results_handler.final_results["nat_sum_data"].set_index("estimand").to_dict(orient="index")
+        if APP_ENV != "local" and self.save_results:
+            self.results_handler.write_data(
+                self.election_id, self.office, self.geographic_unit_type, keys=["nat_sum_data"]
             )
-            nat_sum_estimates = {
-                e: [
-                    nat_sum_estimates[e]["agg_pred"],
-                    nat_sum_estimates[e]["agg_lower"],
-                    nat_sum_estimates[e]["agg_upper"],
-                ]
-                for e in nat_sum_estimates
-            }
 
         return nat_sum_estimates
 
@@ -230,19 +227,12 @@ class ModelClient:
         fixed_effects = kwargs.get("fixed_effects", {})
         pi_method = kwargs.get("pi_method", "nonparametric")
         save_output = kwargs.get("save_output", ["results"])
-        save_results = "results" in save_output
+        self.save_results = "results" in save_output
         save_data = "data" in save_output
         save_config = "config" in save_output
         # saving conformalization data only makes sense if a ConformalElectionModel is used
         save_conformalization = "conformalization" in save_output
         handle_unreporting = kwargs.get("handle_unreporting", "drop")
-
-        national_summary = kwargs.get("national_summary", False)
-        if national_summary:
-            nat_sum_data_dict = kwargs.get("nat_sum_data_dict", None)
-            called_states = kwargs.get("called_states", None)
-            base_to_add = kwargs.get("base_to_add", 0)
-            alpha = kwargs.get("alpha", 0.99)
 
         district_election = False
         if office in {"H", "Y", "Z"}:
@@ -276,15 +266,18 @@ class ModelClient:
             model_parameters,
             handle_unreporting,
         )
+        self.election_id = election_id
+        self.office = office
+        self.geographic_unit_type = geographic_unit_type
 
-        states_with_election = config_handler.get_states(office)
-        estimand_baselines = config_handler.get_estimand_baselines(office, estimands)
+        states_with_election = config_handler.get_states(self.office)
+        estimand_baselines = config_handler.get_estimand_baselines(self.office, estimands)
 
-        LOG.info("Getting preprocessed data: %s", election_id)
+        LOG.info("Getting preprocessed data: %s", self.election_id)
         preprocessed_data_handler = PreprocessedDataHandler(
-            election_id,
-            office,
-            geographic_unit_type,
+            self.election_id,
+            self.office,
+            self.geographic_unit_type,
             estimands,
             estimand_baselines,
             data=preprocessed_data,
@@ -302,7 +295,7 @@ class ModelClient:
             preprocessed_data,
             current_data,
             estimands,
-            geographic_unit_type,
+            self.geographic_unit_type,
             handle_unreporting=handle_unreporting,
         )
 
@@ -335,7 +328,6 @@ class ModelClient:
             self.model = GaussianElectionModel(model_settings=model_settings)
         elif pi_method == "bootstrap":
             self.model = BootstrapElectionModel(model_settings=model_settings)
-        self._model_updated = True
 
         minimum_reporting_units_max = 0
         for alpha in prediction_intervals:
@@ -343,8 +335,8 @@ class ModelClient:
             if minimum_reporting_units > minimum_reporting_units_max:
                 minimum_reporting_units_max = minimum_reporting_units
 
-        if APP_ENV != "local" and save_results:
-            data.write_data(election_id, office)
+        if APP_ENV != "local" and self.save_results:
+            data.write_data(self.election_id, self.office)
 
         n_reporting_expected_units = reporting_units.shape[0]
         n_unexpected_units = unexpected_units.shape[0]
@@ -389,7 +381,7 @@ class ModelClient:
             self.results_handler.add_unit_intervals(estimand, alpha_to_unit_prediction_intervals)
 
             for aggregate in self.results_handler.aggregates:
-                aggregate_list = self.get_aggregate_list(office, aggregate)
+                aggregate_list = self.get_aggregate_list(self.office, aggregate)
                 estimates_df = self.model.get_aggregate_predictions(
                     self.results_handler.reporting_units,
                     self.results_handler.nonreporting_units,
@@ -420,11 +412,8 @@ class ModelClient:
 
         self.results_handler.process_final_results()
 
-        if national_summary:
-            self.get_national_summary_votes_estimates(nat_sum_data_dict, called_states, base_to_add, alpha)
-
-        if APP_ENV != "local" and save_results:
-            self.results_handler.write_data(election_id, office, geographic_unit_type)
+        if APP_ENV != "local" and self.save_results:
+            self.results_handler.write_data(self.election_id, self.office, self.geographic_unit_type)
 
         return self.results_handler.final_results
 
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -207,6 +207,7 @@ class ModelClient:
         # saving conformalization data only makes sense if a ConformalElectionModel is used
         save_conformalization = "conformalization" in save_output
         handle_unreporting = kwargs.get("handle_unreporting", "drop")
+        national_summary = kwargs.get("national_summary", False)
 
         district_election = False
         if office in {"H", "Y", "Z"}:
@@ -382,6 +383,12 @@ class ModelClient:
                 )
 
         results_handler.process_final_results()
+
+        if national_summary:
+            # TODO: parameters for self.get_national_summary_votes_estimates()
+            nat_sum_estimates = self.get_national_summary_votes_estimates(nat_sum_data_dict=None, called_states=None)
+            results_handler.add_national_summary_estimates(nat_sum_estimates)
+
         if APP_ENV != "local" and save_results:
             results_handler.write_data(election_id, office, geographic_unit_type)
 
--- a/src/elexmodel/client.py
+++ b/src/elexmodel/client.py
@@ -41,6 +41,7 @@ class ModelClient:
         self.all_conformalization_data_unit_dict = defaultdict(dict)
         self.all_conformalization_data_agg_dict = defaultdict(dict)
         self.model = None
+        self.results_handler = None
 
     def _check_input_parameters(
         self,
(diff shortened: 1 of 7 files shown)


lennybronner (commented):
I'm not opposed to moving the call for `self.model.get_national_summary_estimates(None, None, 0, 0.99)` to within `get_estimates`. At the moment it is executed outside the model which then calls `get_national_summary_votes_estimates` in the client. But if we want to make this change we need to:
1. Talk to the live team, because this changes how they call the aggregate model
2. Make corresponding changes in the testbed
3. Make sure we pass through the relevant arguments.

dmnapolitano:
> I'm not opposed to moving the call for `self.model.get_national_summary_estimates(None, None, 0, 0.99)` to within `get_estimates`. At the moment it is executed outside the model which then calls `get_national_summary_votes_estimates` in the client. But if we want to make this change we need to:
> 
> 1. Talk to the live team, because this changes how they call the aggregate model
> 2. Make corresponding changes in the testbed
> 3. Make sure we pass through the relevant arguments.

Yes!  This is great, I was wondering about all of this 🎉   I moved it because `self.model` is instantiated in `get_estimates()` so if one called `get_national_summary_votes_estimates()` first, it would fail, and I wasn't sure there was a scenario where we would ever want call `get_national_summary_votes_estimates()` without having called `get_estimates()` 🤔   So this is good.  I'll check with the team when Jen is online and see what they say 😄 👍🏻

dmnapolitano:
Also I just realized this is PR 100 😄 🎉 💯

dmnapolitano:
> > I'm not opposed to moving the call for `self.model.get_national_summary_estimates(None, None, 0, 0.99)` to within `get_estimates`. At the moment it is executed outside the model which then calls `get_national_summary_votes_estimates` in the client. But if we want to make this change we need to:
> > 
> > 1. Talk to the live team, because this changes how they call the aggregate model
> > 2. Make corresponding changes in the testbed
> > 3. Make sure we pass through the relevant arguments.
> 
> Yes! This is great, I was wondering about all of this 🎉 I moved it because `self.model` is instantiated in `get_estimates()` so if one called `get_national_summary_votes_estimates()` first, it would fail, and I wasn't sure there was a scenario where we would ever want to call `get_national_summary_votes_estimates()` without having called `get_estimates()` 🤔 So this is good. I'll check with the team when Jen is online and see what they say 😄 👍🏻

Alright so the feedback from Jen is please don't change anything 😂  So I'll add the method back in and I have some ideas for how to accomplish this... 🤔

dmnapolitano:
> > > I'm not opposed to moving the call for `self.model.get_national_summary_estimates(None, None, 0, 0.99)` to within `get_estimates`. At the moment it is executed outside the model which then calls `get_national_summary_votes_estimates` in the client. But if we want to make this change we need to:
> > > 
> > > 1. Talk to the live team, because this changes how they call the aggregate model
> > > 2. Make corresponding changes in the testbed
> > > 3. Make sure we pass through the relevant arguments.
> > 
> > 
> > Yes! This is great, I was wondering about all of this 🎉 I moved it because `self.model` is instantiated in `get_estimates()` so if one called `get_national_summary_votes_estimates()` first, it would fail, and I wasn't sure there was a scenario where we would ever want to call `get_national_summary_votes_estimates()` without having called `get_estimates()` 🤔 So this is good. I'll check with the team when Jen is online and see what they say 😄 👍🏻
> 
> Alright so the feedback from Jen is please don't change anything 😂 So I'll add the method back in and I have some ideas for how to accomplish this... 🤔

@lennybronner alright I just pushed some changes that I think should accomplish this and preserve the current `get_national_summary_votes_estimates()` 🙏🏻  Note that the writing of national summary estimates to s3 still only happens in `get_estimates()` and not in `get_national_summary_votes_estimates()`, which might be fine given the behavior of `get_national_summary_votes_estimates()` that live/site/elections is expecting 🤔

src/elexmodel/client.py
```diff
@@ -169,8 +170,33 @@ def get_aggregate_list(self, office, aggregate):
...
-        return self.model.get_national_summary_estimates(nat_sum_data_dict, called_states, base_to_add, alpha)
+    def get_national_summary_votes_estimates(
+        self, nat_sum_data_dict=None, called_states=None, base_to_add=0, alpha=0.99
+    ):
+        if self.model is None:
+            raise ModelClientException(
+                "Must call the get_estimands() method before get_national_summary_votes_estimates()."
+            )
+
+        if "nat_sum_data" in self.results_handler.final_results:
```
lennybronner, line 181: After calling `get_national_summary_votes_estimates` once this would always return the same national summary estimates, rather than creating new ones with the new model updates.

lennybronner (commented):
I'm not sure I understand why so much change was necessary? Why can't we just write the data to s3 from `get_national_summary_votes_estimates`?

src/elexmodel/client.py
dmnapolitano: That's right.  The current way this operates and the code is written, there is no `self.model` without calling `get_estimates()`.  That flow is maintained here.  There's no reason to re-compute the national summary estimates without calling `get_estimates()` first, so might as well return the ones using the model from the previous `get_estimates()` call. 🤷🏻‍♀️

dmnapolitano: I mean, unless there is a reason to re-compute them, in which case I'll take this out lol 🤔

dmnapolitano:
> I'm not sure I understand why so much change was necessary? Why can't we just write the data to s3 from `get_national_summary_votes_estimates`?

We could, but,

1. Is that something we want to happen every time this method is called?  Is that a side effect the live team is expecting?
2. We'll have to add additional arguments to `get_national_summary_votes_estimates()` to handle the logic of where the file gets written to (could be optional though);
3. I'd have to write something else to write this out to s3, rather than reuse the code writing out other results files;
4. I had no way of testing this without running the testbed or adding the CLI argument (which you are also adding in PR #101 🙌🏻 );

I think those were my main thoughts on this.  I'll take a look again to see what can be simplified, but, 🤔

src/elexmodel/client.py
lennybronner: The national summary prediction might change any time the model is run

lennybronner:
> > I'm not sure I understand why so much change was necessary? Why can't we just write the data to s3 from `get_national_summary_votes_estimates`?
> 
> We could, but,
> 
> 1. Is that something we want to happen every time this method is called?  Is that a side effect the live team is expecting?
> 2. We'll have to add additional arguments to `get_national_summary_votes_estimates()` to handle the logic of where the file gets written to (could be optional though);
> 3. I'd have to write something else to write this out to s3, rather than reuse the code writing out other results files;
> 4. I had no way of testing this without running the testbed or adding the CLI argument (which you are also adding in PR [Elex 4455 add race call contest agg #101](https://github.com/washingtonpost/elex-live-model/pull/101) 🙌🏻 );
> 
> I think those were my main thoughts on this. I'll take a look again to see what can be simplified, but, 🤔

I don't really like re-running the national summary estimates, I'm not sure we've thought about whether something changes. I understand why you just return the national summary estimates as they exist now, but that only runs the aggregate model once per instantiation -- which we do not want. Maybe we add a parameter to `get_national_summary_votes_estimates` that tells us whether we should re-run or just return the current version.

dmnapolitano:
> > > I'm not sure I understand why so much change was necessary? Why can't we just write the data to s3 from `get_national_summary_votes_estimates`?
> > 
> > 
> > We could, but,
> > 
> > 1. Is that something we want to happen every time this method is called?  Is that a side effect the live team is expecting?
> > 2. We'll have to add additional arguments to `get_national_summary_votes_estimates()` to handle the logic of where the file gets written to (could be optional though);
> > 3. I'd have to write something else to write this out to s3, rather than reuse the code writing out other results files;
> > 4. I had no way of testing this without running the testbed or adding the CLI argument (which you are also adding in PR [Elex 4455 add race call contest agg #101](https://github.com/washingtonpost/elex-live-model/pull/101) 🙌🏻 );
> > 
> > I think those were my main thoughts on this. I'll take a look again to see what can be simplified, but, 🤔
> 
> I don't really like re-running the national summary estimates, I'm not sure we've thought about whether something changes. I understand why you just return the national summary estimates as they exist now, but that only runs the aggregate model once per instantiation -- which we do not want. Maybe we add a parameter to `get_national_summary_votes_estimates` that tells us whether we should re-run or just return the current version.

Ok, I just pushed a possible solution for this based on this comment and the one you have on the code 🙌🏻  The thinking is this:

1. The model only gets instantiated/updated/re-run when `get_estimates()` is called;
2. `get_national_summary_votes_estimates()` requires a model, so without having called `get_estimates()` first, it will fail;
3. But, if the flow is, `get_estimates()`, then `get_national_summary_votes_estimates()`, then `get_national_summary_votes_estimates()` again, the model hasn't changed so there's no reason to re-compute the national summary votes;
4. So, if we set a flag (`_model_updated`) to `True` whenever `get_estimates()` has been called and `False` otherwise, then check the flag, that should take care of scenario # 3 🎉 

The other option is to re-compute the national summary votes every time `get_national_summary_votes_estimates()`.  If that's a pretty fast thing to do, it might be worth it compared to this logic 😅 🤔

dmnapolitano:
> > > > I'm not sure I understand why so much change was necessary? Why can't we just write the data to s3 from `get_national_summary_votes_estimates`?
> > > 
> > > 
> > > We could, but,
> > > 
> > > 1. Is that something we want to happen every time this method is called?  Is that a side effect the live team is expecting?
> > > 2. We'll have to add additional arguments to `get_national_summary_votes_estimates()` to handle the logic of where the file gets written to (could be optional though);
> > > 3. I'd have to write something else to write this out to s3, rather than reuse the code writing out other results files;
> > > 4. I had no way of testing this without running the testbed or adding the CLI argument (which you are also adding in PR [Elex 4455 add race call contest agg #101](https://github.com/washingtonpost/elex-live-model/pull/101) 🙌🏻 );
> > > 
> > > I think those were my main thoughts on this. I'll take a look again to see what can be simplified, but, 🤔
> > 
> > 
> > I don't really like re-running the national summary estimates, I'm not sure we've thought about whether something changes. I understand why you just return the national summary estimates as they exist now, but that only runs the aggregate model once per instantiation -- which we do not want. Maybe we add a parameter to `get_national_summary_votes_estimates` that tells us whether we should re-run or just return the current version.
> 
> Ok, I just pushed a possible solution for this based on this comment and the one you have on the code 🙌🏻 The thinking is this:
> 
> 1. The model only gets instantiated/updated/re-run when `get_estimates()` is called;
> 2. `get_national_summary_votes_estimates()` requires a model, so without having called `get_estimates()` first, it will fail;
> 3. But, if the flow is, `get_estimates()`, then `get_national_summary_votes_estimates()`, then `get_national_summary_votes_estimates()` again, the model hasn't changed so there's no reason to re-compute the national summary votes;
> 4. So, if we set a flag (`_model_updated`) to `True` whenever `get_estimates()` has been called and `False` otherwise, then check the flag, that should take care of scenario # 3 🎉
> 
> The other option is to re-compute the national summary votes every time `get_national_summary_votes_estimates()` is called. If that's a pretty fast thing to do, it might be worth it compared to this logic 😅 🤔

Although either way, I just realized if you call that method and you're calling `get_estimates()` with `national_summary = False`, those updated national summary estimates won't get written to s3 🤦🏻‍♀️  Blargh, sorry, let me continue to think about this.... 🤔

dmnapolitano:
Ok @lennybronner take a look at the changes I just pushed.  It borrows some logic / variables from `get_estimands()` to write specifically the national summary estimates to s3.  The only thing that doesn't happen here is, `self.model.get_national_summary_estimates()` is re-called every time, which might be fine since this seems to be a pretty fast method, although I'd be curious to hear your and live/elections's experience with that 😬  If it ends up being too slow, I can add the logic bad in to avoid re-computing the national summary estimates.  Anyway, thanks and LMK what you think about this approach 🤔

lennybronner:
This generally looks good now, thanks so much for figuring this out. I left one small question. This will necessitate making changes to the model testbed, so please make those (and then run the 2020 election through the testbed to make sure it all still works as expected). Also, the live will need to make small changes to how they interact with the national summary data, so as long as they aware aware of this, this seems fine.

Where does re-calling `get_national_summary_estimates` happen? I only see it happen once, where it should be.

dmnapolitano:
> This generally looks good now, thanks so much for figuring this out. I left one small question. This will necessitate making changes to the model testbed, so please make those (and then run the 2020 election through the testbed to make sure it all still works as expected). Also, the live will need to make small changes to how they interact with the national summary data, so as long as they aware aware of this, this seems fine.
> 
> Where does re-calling `get_national_summary_estimates` happen? I only see it happen once, where it should be.

Thanks! 🎉 

1. What are the changes that need to be made to the model testbed? 🤔 
2. What are the changes that the live team will need to make?  I tried to make it so they don't have to change anything, but maybe I failed somehow 🤔 
3. Sorry, I explained that poorly.  If you do (1) `get_estimates()`, (2) `get_national_summary_votes_estimates()`, then (3) `get_national_summary_votes_estimates()`, that second call (3) will re-compute the national vote estimates and return the same results as (2).  I was trying to check for this before but I took it out because it's a bit complicated and might not be worth the extra effort.  So I figured I'd revisit it if we got complaints about speed lol 😄

src/elexmodel/handlers/data/ModelResults.py
```diff
@@ -115,6 +122,8 @@ def write_data(self, election_id, office, geographic_unit_type):
             self.process_final_results()
         s3_client = s3.S3CsvUtil(TARGET_BUCKET)
         for key, value in self.final_results.items():
+            if keys is not None and key not in keys:
```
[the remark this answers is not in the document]
dmnapolitano: If you pass in `keys` like `keys=["nat_sum_data"]`, then only the `nat_sum_data/current.csv` will be written.  It's to prevent every single CSV being written every single time this method is called, since we only want `nat_sum_data/current.csv` to be updated when `get_national_summary_votes_estimates()` is called 🤔

lennybronner:
> > This generally looks good now, thanks so much for figuring this out. I left one small question. This will necessitate making changes to the model testbed, so please make those (and then run the 2020 election through the testbed to make sure it all still works as expected). Also, the live will need to make small changes to how they interact with the national summary data, so as long as they aware aware of this, this seems fine.
> > Where does re-calling `get_national_summary_estimates` happen? I only see it happen once, where it should be.
> 
> Thanks! 🎉
> 
> 1. What are the changes that need to be made to the model testbed? 🤔
> 2. What are the changes that the live team will need to make?  I tried to make it so they don't have to change anything, but maybe I failed somehow 🤔
> 3. Sorry, I explained that poorly.  If you do (1) `get_estimates()`, (2) `get_national_summary_votes_estimates()`, then (3) `get_national_summary_votes_estimates()`, that second call (3) will re-compute the national vote estimates and return the same results as (2).  I was trying to check for this before but I took it out because it's a bit complicated and might not be worth the extra effort.  So I figured I'd revisit it if we got complaints about speed lol 😄

(1) and (2) The structure of the response of `get_national_summary_votes_estimates` changed, it's not a dataframe rather than a list, right?

(3) that's what we want.

dmnapolitano:
> > > This generally looks good now, thanks so much for figuring this out. I left one small question. This will necessitate making changes to the model testbed, so please make those (and then run the 2020 election through the testbed to make sure it all still works as expected). Also, the live will need to make small changes to how they interact with the national summary data, so as long as they aware aware of this, this seems fine.
> > > Where does re-calling `get_national_summary_estimates` happen? I only see it happen once, where it should be.
> > 
> > 
> > Thanks! 🎉
> > 
> > 1. What are the changes that need to be made to the model testbed? 🤔
> > 2. What are the changes that the live team will need to make?  I tried to make it so they don't have to change anything, but maybe I failed somehow 🤔
> > 3. Sorry, I explained that poorly.  If you do (1) `get_estimates()`, (2) `get_national_summary_votes_estimates()`, then (3) `get_national_summary_votes_estimates()`, that second call (3) will re-compute the national vote estimates and return the same results as (2).  I was trying to check for this before but I took it out because it's a bit complicated and might not be worth the extra effort.  So I figured I'd revisit it if we got complaints about speed lol 😄
> 
> (1) and (2) The structure of the response of `get_national_summary_votes_estimates` changed, it's not a dataframe rather than a list, right?
> 
> (3) that's what we want.

(3) Great! 🎉 

(1) and (2), no, `get_national_summary_votes_estimates()` in `ModelClient` returned something like this:

```python
{"margin" : [agg_pred, agg_lower, agg_upper]}
```

I made sure to preserve that.  When writing to s3, I'm currently converting that to a CSV but that can easily be a JSON or whatever is best 🤔

lennybronner:
amazing, thank you! And just to confirm, you've run this from the testbed (including writing to s3) and it works?

dmnapolitano:
> amazing, thank you! And just to confirm, you've run this from the testbed (including writing to s3) and it works?

Thanks!!  I will test that now! 🤔

dmnapolitano:
> > amazing, thank you! And just to confirm, you've run this from the testbed (including writing to s3) and it works?
> 
> Thanks!! I will test that now! 🤔

@lennybronner ok, so this testbed command works as we'd expect it to:

```
python run.py 2022-11-08_USA_G redo --office_id S_county --fixed_effects "['county_classification']" --geographic_unit_type county --pi_method bootstrap --estimands "['margin']" --features "['baseline_normalized_margin', 'education_bachelors_or_higher']" --end_timestamp "2022-11-10 22:00:29-05:00" --agg_model_preds
```

🎉 but for writing to s3 from the testbed, there's no way to do that via the CLI currently, although I could add the options.  And if I did, it would just write through `ModelClient` and then overwrite every set of predictions produced.  Is that ok or would writing everything we save in `output/` be better? 🤔

lennybronner:
Yes, this is why I meant this needs a corresponding PR in the testbed. It's ok to overwrite the predictions produced in dev in s3.

dmnapolitano:
> > Yes, this is why I meant this needs a corresponding PR in the testbed. It's ok to overwrite the predictions produced in dev in s3.
> 
> Oh sorry, I misunderstood! Ok, I'll work on that now 😅 👍🏻

@lennybronner this is ready for review 🎉  washpost/elex-live-model-testbed#25

dmnapolitano:
> Great, and now just to confirm, you ran this with your testbed branch?

Yup! 🎉

lennybronner (approved):

PR #101, opened 2024-09-04
Elex 4455 add race call contest agg

lennybronner:
## Description
We move the race calling from `get_national_summary` to `get_aggregate_predictions` and `get_aggregate_prediction_intervals`. This means that we apply the race calls to the contest level predictions also (ie. state predictions or house district predictions) instead of purely on the national summary predictions. 

This is more conceptually correct, but will also let us show the our contest level predictions after race calls have already happened if we want.

Unit tests were also updated to take this into account.

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-4455

## Test Steps
```
elexmodel 2017-11-07_VA_G --estimands=margin --office_id=G --geographic_unit_type=county --pi_method bootstrap --features baseline_normalized_margin --called_contests='{"VA": 0}' --percent_reporting 10 --aggregates postal_code --aggregates county_classification
```
This now allows a force call for Republicans (and you can see the predictions and intervals all be below zero for the state as a result). 

If you want to test on aggregate model, then you can just add more contests using this dictionary (either in the CLI or using the testbed).
```
{'VA-47': 0, 'VA-50': 0, 'VA-51': 0, 'VA-53': 0, 'VA-55': 1, 'VA-61': 0, 'VA-62': 0, 'VA-89': 0, 'VA-64': 0, 'VA-65': 1, 'VA-66': 0, 'VA-70': 1, 'VA-71': 0, 'VA-73': 0, 'VA-75': 0, 'VA-82': -1, 'VA-60': 0, 'VA-76': 1, 'VA-83': 0, 'VA-84': 1, 'VA-41': -1, 'VA-33': 0, 'VA-86': 0, 'VA-87': 1, 'VA-88': 1, 'VA-90': 0, 'VA-93': 1, 'VA-94': 1, 'VA-96': 1, 'VA-34': 0, 'VA-97': 1, 'VA-99': 0, 'VA-72': 0, 'VA-31': 0, 'VA-11': 1, 'VA-30': 0, 'VA-52': 0, 'VA-91': 1, 'VA-92': 1, 'VA-100': 0, 'VA-6': 1, 'VA-7': 1, 'VA-8': 1, 'VA-9': 1, 'VA-10': 1, 'VA-14': 1, 'VA-15': 1, 'VA-18': 1, 'VA-21': -1, 'VA-32': 0, 'VA-36': 0, 'VA-37': 0, 'VA-40': 0, 'VA-57': 0, 'VA-58': 1, 'VA-59': 0, 'VA-23': 1, 'VA-20': 1, 'VA-22': 0, 'VA-25': 1, 'VA-26': 1, 'VA-27': 1, 'VA-28': 1, 'VA-29': 1, 'VA-67': 0}
```

also `pytest`

### DIFF
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -1061,11 +1061,10 @@ class BootstrapElectionModel(BaseElectionModel):
             len(aggregate) == 2 and "postal_code" in aggregate and "district" in aggregate
         )
 
-    def _call_contest(self, to_call: np.array, called_contests: dict) -> np.array:
+    def _format_called_contests_dictionary(self, called_contests: dict | None) -> dict:
         """
-        This functions applies race calls to the point prediction
+        Make sure that the called contest dictionary has the correct format
         """
-
         # called_contests is a dictionary where 1 means that the LHS party has won, 0 means that the RHS party has won
         # and -1 means that the contest is not called. If called_contests is None, assume that all contests are not called.
         if called_contests is None or len(called_contests) == 0:
@@ -1075,6 +1074,21 @@ class BootstrapElectionModel(BaseElectionModel):
             raise BootstrapElectionModelException(
                 f"called_contests is of length {len(called_contests)} but there are {self.n_contests} contests"
             )
+
+        called_contest_acceptable_values = {0, 1, -1}
+        if not all(value in called_contest_acceptable_values for value in called_contests.values()):
+            raise BootstrapElectionModelException(
+                f"called_contest values need to be either 0, 1, or -1. But current value is {called_contest}"
+            )
+
+        return called_contests
+
+    def _call_contest(self, to_call: np.array, called_contests: dict) -> np.array:
+        """
+        This functions applies race calls to the point prediction
+        """
+
+        called_contests = self._format_called_contests_dictionary(called_contests)
 
         # array sorted by contest, where the element is the call indicator (1, 0 or -1)
         contest_call_array =  np.array([contest_tuple[1] for contest_tuple in sorted(called_contests.items())])
@@ -1327,7 +1341,7 @@ class BootstrapElectionModel(BaseElectionModel):
         if self._is_top_level_aggregate(aggregate):
             aggregate_perc_margin_total = self.aggregate_pred_margin
 
-            called_contests = kwargs.get("called_contests")
+            called_contests = self._format_called_contests_dictionary(kwargs.get("called_contests", {}))
             interval_upper, interval_lower = (aggregate_perc_margin_total - np.quantile(error_diff, q=[lower_q, upper_q], axis=-1).T).T
 
 
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -1086,7 +1086,7 @@ class BootstrapElectionModel(BaseElectionModel):
 
         return called_contests
 
-    def _call_contest(self, to_call: np.array, called_contests: dict) -> np.array:
+    def _adjust_called_contests(self, to_call: np.array, called_contests: dict) -> np.array:
         """
         This functions applies race calls to the point prediction
         """
@@ -1182,7 +1182,9 @@ class BootstrapElectionModel(BaseElectionModel):
         # which we will need for the national summary (e.g. ecv) model
         if self._is_top_level_aggregate(aggregate):
             called_contests = kwargs.get("called_contests")
-            self.aggregate_pred_margin = self._call_contest(raw_margin_df.pred_margin, called_contests).reshape(-1, 1)
+            self.aggregate_pred_margin = self._adjust_called_contests(
+                raw_margin_df.pred_margin, called_contests
+            ).reshape(-1, 1)
             raw_margin_df["pred_margin"] = self.aggregate_pred_margin
 
             aggregate_sum = all_units.groupby(aggregate_temp_column_name).sum()
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -1321,12 +1321,13 @@ class BootstrapElectionModel(BaseElectionModel):
 
         lower_q, upper_q = self._get_quantiles(alpha)
 
+        error_diff = divided_error_B_1 - divided_error_B_2
+
         # saves the aggregate errors in case we want to generate somem form of national predictions (like ecv)
         if self._is_top_level_aggregate(aggregate):
             aggregate_perc_margin_total = self.aggregate_pred_margin
 
             called_contests = kwargs.get("called_contests")
-            error_diff = divided_error_B_1 - divided_error_B_2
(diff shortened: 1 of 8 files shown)


src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -1054,13 +1061,64 @@ def _is_top_level_aggregate(self, aggregate: list) -> bool:
...
+            in called_contest_acceptable_values
+            for value in called_contests.values()
+        ):
+            raise BootstrapElectionModelException(
+                f"called_contest values need to be either 0, 1, or -1. But current value is {called_contests}"
+            )
+
+        return called_contests
+
+    def _call_contest(self, to_call: np.array, called_contests: dict) -> np.array:
```
dmnapolitano, line 1090: Does `to_call` always contain uncalled contests, i.e. contests not yet called by the AP or by our team?  I'm trying to understand how contests that have already been called are used here.  Does our model re-call (or un-call) called contests?  Do we return all contests and their call statuses (called by model, called by others, uncalled)?  What if a user wanted to know who called the each contest? 🤔

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -1054,13 +1061,64 @@ def _is_top_level_aggregate(self, aggregate: list) -> bool:
...
+            called_contests = {i: -1 for i in range(self.n_contests)}
+
+        if len(called_contests) != self.n_contests:
+            raise BootstrapElectionModelException(
+                f"called_contests is of length {len(called_contests)} but there are {self.n_contests} contests"
+            )
+
+        called_contest_acceptable_values = {0, 1, -1}
+        if not all(
+            any(math.isclose(value, acceptable_value) for acceptable_value in called_contest_acceptable_values)
```
dmnapolitano, line 1080: What's the difference between using `math.isclose()` instead of `np.isclose()`? 🤔

tests/models/test_bootstrap_election_model.py
```diff
@@ -579,6 +579,72 @@ def test_is_top_level_aggregate(bootstrap_election_model):
     assert not bootstrap_election_model._is_top_level_aggregate([])
 
 
+def test_format_called_contests_dictionary(bootstrap_election_model, rng):
+    n_contests = 10
+    bootstrap_election_model.n_contests = n_contests
```
dmnapolitano, line 584: I've noticed this in a bunch of places.  Why do we have to set this here instead of getting it from `aggregate_indicator.shape[1]` like we do in the `BootstrapElectionModel` class? 🤔

src/elexmodel/models/BootstrapElectionModel.py
lennybronner: I'm not sure I understand your question. `to_call` is a vector of the predicted margin in all races, whether called or not. Our model does not call races at all, this function forces the vector of predictions to be on the correct side of zero when the call happens (ie. we are predicting GOP wins by 0.01% but the race is called for dems, this forces the prediction to be on the dem side).

The call status is given to the model by the user, so no need to return it.

tests/models/test_bootstrap_election_model.py
lennybronner: because the aggregate indicator does not exist in the unit test, since we are never forcing the model to run the bootstrap.

src/elexmodel/models/BootstrapElectionModel.py
dmnapolitano: In that case, how would you feel about replacing it with `np.isclose()`?  We're already importing `numpy` and if there's no real difference, I don't see how it's worth the extra memory to import `math` for that one function 🤔

tests/models/test_bootstrap_election_model.py
dmnapolitano: Ok, I see, so for testing purposes we want to bypass forcing the model to run the bootstrap so we can test this specific functionality.  I was just curious how common this could be and if we should add something to the `BootstrapElectionModel` class, but for testing, this is good 😄 🎉

src/elexmodel/models/BootstrapElectionModel.py
dmnapolitano: Ok, I think I follow.  So this method's purpose is to inform our model of the races (contests) that have been called, is that right?  I thought the idea of this method was to determine which races, based on our model's predictions/PIs, were in call status (not necessarily to call the race itself but to return such a suggestion to the user), which is why I wasn't sure what we were doing with races that had already been called 😓 

If I have any of that right now 😬 how would you feel about renaming this method to `_update_call_status()` or something similar? 🤔

lennybronner: Specifically updating the point prediction (the prediction intervals get updated in the function specific to the aggregate prediction intervals). I don't love that name, call status is not really a phrase we've used for the model before. How about `adjust_predictions`?

src/elexmodel/models/BootstrapElectionModel.py
lennybronner: done!

tests/models/test_bootstrap_election_model.py
lennybronner: It is actually pretty common! What do you think we could add there? 👀

src/elexmodel/models/BootstrapElectionModel.py
dmnapolitano: Sure! 🎉   Or maybe `_adjust_predictions_based_on_race_calls()`?  Hmm that's a lot of letters.  `_adjust_called_contests()`? 🤔

tests/models/test_bootstrap_election_model.py
dmnapolitano: Oh wow!  Ok, in that case, what about this:

```python
class BootstrapElectionModel:
    def __init__(self, <current arguments>, n_contests=None): # or whatever n_contests default value you think is best
      # other code etc., then
       if aggregate_indicator is None:
          self.n_contests = n_contests
      else:
          self.n_contests = aggregate_indicator.shape[1]
```

I think that should work 🤔

dmnapolitano: Although, in these unit tests `BootstrapElectionModel` is already instantiated so that doesn't help you 😢  You could also add a setter:

```python
# add self.n_contests to def __init__(), then

@n_contests.setter
def n_contests(self, n):
    # you can do some verification on `n` if you wanted to here
    self.n_contests = n
```

But that really doesn't give you anything special above what you have right now, other than people can't get `n_contests` without you adding a getter, although it does allow you to add in some verification if you needed it 🤔 🤔

lennybronner: hmm in that case lets leave it for now (but thanks for looking into this!)

src/elexmodel/models/BootstrapElectionModel.py
lennybronner: done!

dmnapolitano (approved):
Thanks for answering all my weird questions! 😄 🎉 🎉

jjcherian:
LGTM.

PR #105, opened 2024-09-16
ELEX-3394 Part 1: distinguish between unexpected and non-predictive units

dmnapolitano:
## Description

Hi!  The changes in this PR:

* refactor the way we determine reporting units, non-reporting units, and unexpected units so that `_get_unexpected_units()` is only called once rather than three times (which was once per each of reporting, non-reporting, and unexpected);
* introduce the `predictive` indicator in the dataframe of unexpected units, where `True` means the unit is genuinely unexpected but predictive, and `False` means otherwise (currently above/below the turnout factor thresholds);
* log a dictionary of non-predictive units that looks like `{postal_code : [list of geographic_unit_fips]}`;
* provide unit tests for these

🎉 

## Jira Ticket

[ELEX-3394] although this only resolves the "Thoughts" section of the ticket 😄  (Hence "Part 1")

## Test Steps

`tox` or here's a CLI command that returns one non-predictive unit:

```
elexmodel 2022-11-08_USA_G --estimands=margin --features=baseline_normalized_margin --office_id=G_county --geographic_unit_type=county --pi_method bootstrap --national_summary --save_output results
```

[ELEX-3394]: https://arcpublishing.atlassian.net/browse/ELEX-3394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elexmodel/handlers/data/CombinedData.py
+++ b/src/elexmodel/handlers/data/CombinedData.py
@@ -65,7 +65,7 @@ class CombinedDataHandler:
         )
 
         # identify unexpected and non-predictive units
-        unexpected_units = self._get_unexpected_units()
+        unexpected_units = self._get_unexpected_units(aggregates)
         non_modeled_units = self._get_non_modeled_units(
             percent_reporting_threshold, turnout_factor_lower, turnout_factor_upper
         )
@@ -106,19 +106,7 @@ class CombinedDataHandler:
         # finalize all unexpected/non-modeled units
         unexpected_units["unit_category"] = "unexpected"
         non_modeled_units["unit_category"] = "non-modeled"
-
         all_unexpected_units = pd.concat([unexpected_units, non_modeled_units]).reset_index(drop=True)
-        # since we were not expecting them, we have don't have their county or district
-        # from preprocessed data. so we have to add that back in.
-        if "county_fips" in aggregates:
-            all_unexpected_units["county_fips"] = all_unexpected_units["geographic_unit_fips"].apply(
-                self._get_county_fips_from_geographic_unit_fips
-            )
-        if "district" in aggregates:
-            all_unexpected_units["district"] = all_unexpected_units["geographic_unit_fips"].apply(
-                self._get_district_from_geographic_unit_fips
-            )
-
         all_unexpected_units["reporting"] = int(0)
 
         return (reporting_units, nonreporting_units, all_unexpected_units)
@@ -152,7 +140,7 @@ class CombinedDataHandler:
         components = geographic_unit_fips.split("_")
         return components[0]
 
-    def _get_unexpected_units(self):
+    def _get_unexpected_units(self, aggregates):
         expected_geographic_units = self._get_expected_geographic_unit_fips().tolist()
         no_baseline_units = self._get_units_without_baseline()
         # Note: this uses current_data because self.data drops unexpected units
@@ -165,6 +153,18 @@ class CombinedDataHandler:
             .drop_duplicates(subset="geographic_unit_fips")
             .copy()
         )
+
+        # since we were not expecting them, we have don't have their county or district
+        # from preprocessed data. so we have to add that back in.
+        if "county_fips" in aggregates:
+            unexpected_units["county_fips"] = unexpected_units["geographic_unit_fips"].apply(
+                self._get_county_fips_from_geographic_unit_fips
+            )
+        if "district" in aggregates:
+            unexpected_units["district"] = unexpected_units["geographic_unit_fips"].apply(
+                self._get_district_from_geographic_unit_fips
+            )
+
         return unexpected_units
 
     def _get_non_modeled_units(self, percent_reporting_threshold, turnout_factor_lower, turnout_factor_upper):
@@ -180,7 +180,7 @@ class CombinedDataHandler:
                     (self.data.turnout_factor <= turnout_factor_lower)
                     | (self.data.turnout_factor >= turnout_factor_upper)
                 )
-            ][self.current_data.columns]
+            ]
             .drop_duplicates(subset="geographic_unit_fips")
             .copy()
         )
--- a/src/elexmodel/handlers/data/CombinedData.py
+++ b/src/elexmodel/handlers/data/CombinedData.py
@@ -128,23 +128,47 @@ class CombinedDataHandler:
         return components[0]
 
     # TODO: rename unexpected units to be non-modeled units
-    def _get_unexpected_units(self, percent_reporting_threshold, turnout_factor_lower, turnout_factor_upper):
+    def _get_unexpected_units(
+        self, percent_reporting_threshold, turnout_factor_lower, turnout_factor_upper, distinguish=False
+    ):
         expected_geographic_units = self._get_expected_geographic_unit_fips().tolist()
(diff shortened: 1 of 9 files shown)


src/elexmodel/handlers/data/CombinedData.py
```diff
@@ -74,29 +86,41 @@ def get_reporting_units(self, percent_reporting_threshold, turnout_factor_lower,
...
+        # since we were not expecting them, we have don't have their county or district
+        # from preprocessed data. so we have to add that back in.
+        if "county_fips" in aggregates:
+            all_unexpected_units["county_fips"] = all_unexpected_units["geographic_unit_fips"].apply(
+                self._get_county_fips_from_geographic_unit_fips
+            )
+        if "district" in aggregates:
+            all_unexpected_units["district"] = all_unexpected_units["geographic_unit_fips"].apply(
+                self._get_district_from_geographic_unit_fips
+            )
```
lennybronner, line 118: unless I am mistaken, this should only be necessary for the the truly unexpected units, right? not for the non-modeled units?

src/elexmodel/handlers/data/CombinedData.py
```diff
@@ -74,29 +86,41 @@ def get_reporting_units(self, percent_reporting_threshold, turnout_factor_lower,
...
+
+        all_unexpected_units["reporting"] = int(0)
+        all_unexpected_units["expected"] = False
```
lennybronner, line 121: how about, instead of having multiple strings here that do slightly different things. we have one field called `unit_category` which is either `reporting` or `non_reporting` or `unexpected` or `non_modeled`?

src/elexmodel/handlers/data/CombinedData.py
```diff
@@ -132,48 +156,35 @@ def _get_unexpected_units(self, percent_reporting_threshold, turnout_factor_lowe
...
-        if "county_fips" in aggregates:
-            unexpected_units["county_fips"] = unexpected_units["geographic_unit_fips"].apply(
-                self._get_county_fips_from_geographic_unit_fips
-            )
-
-        if "district" in aggregates:
-            unexpected_units["district"] = unexpected_units["geographic_unit_fips"].apply(
-                self._get_district_from_geographic_unit_fips
-            )
+        units_with_strange_turnout_factor = (
```
lennybronner, line 169: I think it would make sense to split this into two functions. One that actually gets the unexpected units and one that gets the nonmodeled units (ie ones with strange turnout factors etc.)

src/elexmodel/handlers/data/CombinedData.py
lennybronner: I guess we need to confirm that the non-modeled units actually get the preprocessed data attached.

src/elexmodel/handlers/data/CombinedData.py
lennybronner: relatedly, I think we should confirm that this column would also be passed to s3 so that we know what category each unit is in (and maybe also returned by the model)?

lennybronner (commented):
This generally looks great. left a few suggestions/questions

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: I can do that 🎉 but here's a question...this is probably insanely rare, but if there's a unit that is both unexpected _and_ non-predictive, which of the two buckets should the unit be placed in?  Or should it be included in both? 🤔

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: Word, I refactored this so that it would only operate on the truly unexpected units and I had a few unit tests fail, saying these aggregates now contain `nan` for a few units 🤔

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: > how about, instead of having multiple strings here that do slightly different things. we have one field called `unit_category` which is either `reporting` or `non_reporting` or `unexpected` or `non_modeled`?

This sounds good and I just pushed changes for this, so let me know what you think.  I didn't get rid of the `reporting` column, though, because I know that one gets aggregated up to a sum total of units that are reporting within an aggregate (`county_fips`, `postal_code`, etc.) 🤔

src/elexmodel/handlers/data/CombinedData.py
lennybronner: That cannot happen. We can only measure whether a unit is non-predictive if we have a baseline result. If it's unexpected we have no preprocessed data, so it doesn't have a baseline result.

src/elexmodel/handlers/data/CombinedData.py
lennybronner: I assume those unit tests are assuming that non predictive units don't have preprocessed data when we pass it into this function? Or maybe I am wrong from the getgo here

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: Got it!  Ok, in that case I was able to split this into two functions, but I did have to add a couple rules to `_get_non_modeled_units()` so that only expected units and units with baselines could be considered as potential non-modeled units.  This fits with these rules and without this, we have two unit tests that would count the same units as unexpected and non-modeled 👀 🎉

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: No, after making the changes to split the determination of unexpected units from non-modeled units, that should be right.  I think I figured out why that one unit test was failing and then was able to move the `aggregates` part to just `_get_unexpected_units()` after all 😄 🎉

src/elexmodel/handlers/data/CombinedData.py
dmnapolitano: > relatedly, I think we should confirm that this column would also be passed to s3 so that we know what category each unit is in (and maybe also returned by the model)?

Not currently, unless I'm missing it somewhere.  Let me see, I should be able to make that happen 🤔

dmnapolitano: > > relatedly, I think we should confirm that this column would also be passed to s3 so that we know what category each unit is in (and maybe also returned by the model)?
> 
> Not currently, unless I'm missing it somewhere. Let me see, I should be able to make that happen 🤔

Done!!  Take a look; I'm merging the `unit_category` column into the `ModelResults.final_results["unit_data"]`, so LMK if there's a better way to do it 🎉

dmnapolitano:
Hi @lennybronner I just pushed the following changes:

* For units that are expected, regardless of whether or not they're reporting, their `unit_category` is now "expected" (instead of "reporting" or "non-reporting");
* The `_get_units_without_baseline()` method has been renamed to `_get_units_with_baseline_of_zero()` and now units with `baseline_turnout == 0` are considered "non-modeled" units as opposed to "unexpected" units;

🎉

lennybronner (approved):
LGTM

PR #109, opened 2024-09-25
Elex 4549 add model flexibility

lennybronner:
## Description
Instead of being stuck with OLS as our base model for the bootstrap model, this PR gives us flexibility to use different models. We were wed to OLS because we had a quick way of computing the leave-one-out-residual, which we needed for the bootstrap, since the training residual would be biased towards zero. We now use k-fold cross validation to get an estimate for the leave-one-out residual (k-fold residual will be greater than or equal to loo residual, so this is if anything more conservative). We also add the ability to play with OLS vs. QR models. In the future, we may expand the models that are being used here.

This needs [this](https://github.com/washingtonpost/elex-solver/pull/27) branch of elex-solver, which allows us to calculate the k-fold residual. 

## Jira Ticket
https://arcpublishing.atlassian.net/browse/ELEX-4549
## Test Steps
```
elexmodel 2017-11-07_VA_G --estimands=margin --office_id=G --geographic_unit_type=county --pi_method bootstrap  --features baseline_normalized_margin --percent_reporting 30 --aggregates postal_code --aggregates unit --model_parameters '{"model_type": "QR"}'
```
vs 
```
elexmodel 2017-11-07_VA_G --estimands=margin --office_id=G --geographic_unit_type=county --pi_method bootstrap  --features baseline_normalized_margin --percent_reporting 30 --aggregates postal_code --aggregates unit --model_parameters '{"model_type": "OLS"}'
```

### DIFF
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -4,9 +4,9 @@ import logging
 
 import numpy as np
 import pandas as pd
+from elexsolver.LinearSolver import LinearSolver
 from elexsolver.OLSRegressionSolver import OLSRegressionSolver
 from elexsolver.QuantileRegressionSolver import QuantileRegressionSolver
-from elexsolver.LinearSolver import LinearSolver
 from scipy.special import expit
 
 from elexmodel.handlers.data.Featurizer import Featurizer
@@ -132,18 +132,13 @@ class BootstrapElectionModel(BaseElectionModel):
                 w_train = x_y_w_train[:, -1]
                 ols_lambda = OLSRegressionSolver()
                 ols_kwargs = {
-                    'lambda_': lambda_,
-                    'fit_intercept': True,
-                    'regularize_intercept': False,
-                    'n_feat_ignore_reg': 1
+                    "lambda_": lambda_,
+                    "fit_intercept": True,
+                    "regularize_intercept": False,
+                    "n_feat_ignore_reg": 1,
                 }
-                ols_lambda.fit(
-                    x_train,
-                    y_train,
-                    weights=w_train,
-                    **ols_kwargs
-                )
-                y_hat_lambda = ols_lambda.predict(x_test)
+                ols_lambda.fit(x_train, y_train, weights=w_train, **ols_kwargs)
+                ols_lambda.predict(x_test)
                 # error is the weighted sum of squares of the residual between
                 # the actual heldout y and the predicted y on the heldout set
                 errors[i] += np.sum(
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 
 INSTALL_REQUIRES = (
     "click>=8.1",
-    "elex-solver>=2.1.1",
+    "elex-solver @ git+https://github.com/washingtonpost/elex-solver.git@updating-residuals",
```
dmnapolitano, line 8: I changed this temporarily so we can test / run the tests here in github 🎉

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -647,6 +660,22 @@ def _sample_test_errors(
...
+        """
+        if self.model_type == "OLS":
+            model = OLSRegressionSolver()
+            model.fit(x, y, weights=weights, **model_kwargs)
+        elif self.model_type == "QR":
+            model = QuantileRegressionSolver()
+            model.fit(x, y, weights=weights, taus=0.5, **model_kwargs)
+        else:
+            raise BootstrapElectionModelException(
+                f"Please use defined model type 'QR' or 'OLS', input is: {self.model_type}"
```
dmnapolitano, line 675: This is such a polite and helpful error message! 😄 🎉

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -806,36 +835,27 @@ def compute_bootstrap_errors(
...
-            weights=weights_train,
-            lambda_=optimal_lambda_z,
-            fit_intercept=True,
-            regularize_intercept=False,
-            n_feat_ignore_reg=1,
-        )
+        model_kwargs = {"fit_intercept": True, "regularize_intercept": False, "n_feat_ignore_reg": 1}
+        if self.model_type == "OLS":
+            # if we run OLS we use cross validation to find the optimal lambda
+            # TODO: why do we not do this for other models also
```
dmnapolitano, line 843: lol wait, why do we not do this for other models also? 🤔 

Is this a "how can I write code to make this happen" question or a theoretical question (or both)? 🤔

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -845,10 +865,24 @@ def compute_bootstrap_errors(
         #   (it is estiamted as the average error in OLS over all the units in a contest)
         #   delta is the unit level error that is unaccounted for by the contest level effect
         residuals_y, epsilon_y_hat, delta_y_hat = self._estimate_model_errors(
-            ols_y, x_train, y_train, aggregate_indicator_train
+            model_y,
+            x_train,
+            y_train,
+            weights_train,
+            aggregate_indicator_train,
+            K=5,
```
dmnapolitano, line 873: Just curious, how did you pick the default of 5 here (and elsewhere)?  I mean I've done (and seen) 3-fold, 5-fold, 10-fold CV but usually my choice there depends on the number of examples in the data set and how much computing power I have 🤔

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -916,49 +950,42 @@ def compute_bootstrap_errors(
...
+
+        model_y_B = self.fit_model(x_train, y_train_B, weights_train, **model_kwargs_y_B)
+        model_z_B = self.fit_model(x_train, z_train_B, weights_train, **model_kwargs_z_B)
+
+        LOG.info("orig. ols coefficients, normalized margin: \n %s", model_y.coefficients.flatten())
+        LOG.info("boot. ols coefficients, normalized margin: \n %s", model_y_B.coefficients.mean(axis=-1))
 
         # we cannot just apply ols_y_B/old_z_B to the test units because that would be missing
         # our contest level random effect
+        # TODO: DO WE NEVER DO ANYTHING WITH THE OUTPUT
```
dmnapolitano, line 970: 🤔

dmnapolitano (commented):
This looks pretty good! 🎉 🎉  I have some questions here and on the `elex-solver` PR, and I was also wondering if you could add documentation about the new solver options to the README? 🤔 🎉

PR #111, opened 2024-10-10
now allows multiple alphas and returns dataframe

lennybronner:
## Description
This PR allows us to pass multiple alphas to `get_national_summary_votes_estimates` and also changes the return object of the function to be a pandas dataframe where each alpha generates a separate lower and upper column.

This is necessary to avoid overwriting the agg model predictions in s3 when we run the model with 0.7 and 0.9 alphas.

## Test Steps
Can be tested from the testbed. Will put the new testbed branch here.


dmnapolitano (approved):
LGTM and thanks for figuring this out!! 🎉 🎉

PR #115, opened 2024-10-23
if we pass a lambda then no need for cv

lennybronner:
## Description
If we pass a lambda to the model, we don't need to do cross validation to find an optimal lambda. This is the real bottleneck our model has, and should help once we get late into election night.

## Test Steps
You can test this in the testbed like so:
```
python run.py 2020-11-03_USA_G redo --office_id P --estimands="['margin']" --features "['baseline_normalized_margin', 'ethnicity_likely_african_american', 'percent_bachelor_or_higher']" --current_run_name "rerun_2020r" --agg_model_preds --start_timestamp 2020-11-03T19:54:29-05:00 --end_timestamp 2020-11-04T06:07:13-05:00 --redo_run_model_every_n 100 --model_parameters "{'lambda_': 1}"
```

### DIFF
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -811,9 +811,11 @@ class BootstrapElectionModel(BaseElectionModel):
         if self.lambda_ is None:
             optimal_lambda_y = self.cv_lambda(x_train, y_train, np.logspace(-3, 2, 20), weights=weights_train)
             optimal_lambda_z = self.cv_lambda(x_train, z_train, np.logspace(-3, 2, 20), weights=weights_train)
+            LOG.info(f"Optimal lambda for y: {optimal_lambda_y}, Optimal lambda for z: {optimal_lambda_z}")
         else:
             optimal_lambda_y = self.lambda_
             optimal_lambda_z = self.lambda_
+            LOG.info(f"Using user provided lambda: {self.lambda_}")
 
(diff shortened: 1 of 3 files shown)


dmnapolitano (approved):
Thank you so much for this!! 🎉 🎉  Just one question, how to we pick the value for `lambda_`? 😄  Have you found that the CV approach keeps selecting (relatively) the same optimal lambda in the majority of cases? 🤔

src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -807,8 +808,12 @@ def compute_bootstrap_errors(
         )
 
         # we use k-fold cross validation to find the optimal lambda for our OLS regression
-        optimal_lambda_y = self.cv_lambda(x_train, y_train, np.logspace(-3, 2, 20), weights=weights_train)
-        optimal_lambda_z = self.cv_lambda(x_train, z_train, np.logspace(-3, 2, 20), weights=weights_train)
+        if self.lambda_ is None:
```
dmnapolitano, line 811: Sorry I just thought of something.  Could you add some logging here that outputs the values for `optimal_lambda_y` and `optimal_lambda_z`?  That way when we run this during testing we can grab the best `lambda_` from the logs and specify that in the config 😄

lennybronner: good idea, that's done!

PR #116, opened 2024-10-23
epsilon sampling + overlap guarantee in aggregate PIs

jjcherian:
## Description

changes to test epsilon sampling that account for limited signal in states with few units reporting + guaranteed overlap in aggregate PIs

### DIFF
--- a/tests/models/test_bootstrap_election_model.py
+++ b/tests/models/test_bootstrap_election_model.py
@@ -412,23 +412,13 @@ def test_sample_test_epsilon(bootstrap_election_model):
     assert epsilon_y.shape == (aggregate_indicator_test.shape[0], bootstrap_election_model.B)
     assert epsilon_z.shape == (aggregate_indicator_test.shape[0], bootstrap_election_model.B)
 
-    # aggregate 3 has no elements in the training set, and rows 2,3,5 in the test set
-    # are parts of aggregate 3
-    assert np.isclose(epsilon_z[0], 0).all()
-    assert np.isclose(epsilon_z[1], 0).all()
+    # test epsilons should never be zero
+    assert not np.isclose(epsilon_z[0], 0).all()
+    assert not np.isclose(epsilon_z[1], 0).all()
     assert not np.isclose(epsilon_z[2], 0).all()
     assert not np.isclose(epsilon_z[3], 0).all()
-    assert np.isclose(epsilon_z[4], 0).all()
+    assert not np.isclose(epsilon_z[4], 0).all()
     assert not np.isclose(epsilon_z[5], 0).all()
-
-    # testing that if there is only one element epsilon_hat that we return 0
-    epsilon_y, epsilon_z = bootstrap_election_model._sample_test_epsilon(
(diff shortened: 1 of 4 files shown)


tests/models/test_bootstrap_election_model.py
```diff
@@ -965,8 +955,8 @@ def test_get_aggregate_prediction_intervals(bootstrap_election_model, rng):
     assert lower.shape == (6, 1)
     assert upper.shape == (6, 1)
 
-    assert lower[2] == pytest.approx(upper[2])  # since c is fully reporting
-    assert lower[5] == pytest.approx(upper[5])  # since all f units are unexpected
+    assert lower[2] == pytest.approx(upper[2] - 0.002)  # since c is fully reporting
+    assert lower[5] == pytest.approx(upper[5] - 0.002)  # since all f units are unexpected
```
dmnapolitano, line 959: I noticed here and throughout the unit tests there are these adjustments like `- 0.002`, `- 0.001`.  Why are they necessary? 🤔

tests/test_client.py
```diff
@@ -876,5 +876,9 @@ def test_get_national_summary_votes_estimates(model_client, va_governor_county_d
 
     current = model_client.get_national_summary_votes_estimates(None, 0, [0.99])
 
-    pd.testing.assert_frame_equal(current, model_client.results_handler.final_results["nat_sum_data"])
-    pd.testing.assert_frame_equal(expected_df, model_client.results_handler.final_results["nat_sum_data"])
+    pd.testing.assert_frame_equal(
+        current, model_client.results_handler.final_results["nat_sum_data"], check_dtype=False
+    )
+    pd.testing.assert_frame_equal(
+        expected_df, model_client.results_handler.final_results["nat_sum_data"], check_dtype=False
```
dmnapolitano, line 883: So I noticed the addition of `check_dtype=False` here and in every open PR right now, and I can confirm that it's necessary with the latest version of `pandas`.  It seems like `pandas` is now more strict with regard to data types.  Previously you could merge `int` (32 and 64) on `float` (32 and 64) just fine, but now this merge fails due to `agg_pred` being `float64` in the `current` and `expected_df` dataframes and `float64` in `model_client.results_handler.final_results["nat_sum_data"]`.  It should be an easy fix though, but not a huge deal/priority 🤔

tests/models/test_bootstrap_election_model.py
jjcherian: This is because I changed the code for generating lower/upper bounds so that interval_lower and interval_upper are always 0.1% away from the prediction - this is unnecessarily conservative of course when all units are reporting (but the model is never up anyways during that time).

tests/test_client.py
lennybronner: thanks for looking into this!

tests/models/test_bootstrap_election_model.py
dmnapolitano: Ok, I see.  So this test (and the others like it) is to make sure that different is present in the lower/upper bounds.  Ok cool, thanks 😄 👍🏻

lennybronner (approved):

PR #117, opened 2024-10-23
Fit separate model for contest

lennybronner:
## Description
We add a parameter where we can pass in states for which we fit a separate intercept and separate params. ie. this would be useful if a specific state is swinging in a different direction than all the others (ie. Florida in 2022) but don't want to blocklist the entire state because we still want to fit a model for it.

## Test Steps
```
elexmodel 2020-11-03_USA_G --estimands=margin --office_id=P --geographic_unit_type=county --pi_method bootstrap --features baseline_normalized_margin --percent_reporting 20 --model_parameters "{'contests_for_separate_model': ['FL', 'GA']}"
```

### DIFF
--- a/src/elexmodel/handlers/data/Featurizer.py
+++ b/src/elexmodel/handlers/data/Featurizer.py
@@ -88,7 +88,7 @@ class Featurizer:
         # and we zero out the original feature column for those states
         additional_state_features = []
         # if 'BB' in self.states_for_separate_model and 'CC' in self.states_for_separate_model:
-            # import pdb; pdb.set_trace()
+        # import pdb; pdb.set_trace()
         for state in self.states_for_separate_model:
             mask = df.postal_code == state
             for feature in self.features:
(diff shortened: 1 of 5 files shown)


src/elexmodel/handlers/data/Featurizer.py
```diff
@@ -81,6 +83,21 @@ def prepare_data(
         if add_intercept is true an intercept column is added to the features and one fixed effect value is dropped
         """
         df = df.copy()  # create copy so we can do things to the values
+
+        # if a state is in the states for separate model, then we add separate feature columns for that state
+        # and we zero out the original feature column for those states
+        additional_state_features = []
+        # if 'BB' in self.states_for_separate_model and 'CC' in self.states_for_separate_model:
+        # import pdb; pdb.set_trace()
```
dmnapolitano, line 91: Leftover debugging 😄

dmnapolitano (approved):
LGTM!! 🎉 🎉

PR #121, opened 2024-10-27
Some small fixes to 2.2.0 release

lennybronner:
## Description
* Adding a minimum number of units before we start running the outlier detection model.
* Removing fitting state specific intercept if passing state to `states_for_separate_model`
* Making sure that the `baseline_normalized_margin` comes right after the intercept when prepping the data to allow regularization.
* Adding an additional log for coefficients

### DIFF
--- a/src/elexmodel/handlers/data/CombinedData.py
+++ b/src/elexmodel/handlers/data/CombinedData.py
@@ -205,7 +205,9 @@ class CombinedDataHandler:
             fixed_effect for fixed_effect in fixed_effects if fixed_effect in reporting_units.columns
         ]
         featurizer = Featurizer(features=features_to_use, fixed_effects=fixed_effects_to_use)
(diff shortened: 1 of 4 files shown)


src/elexmodel/handlers/data/CombinedData.py
```diff
@@ -48,6 +48,7 @@ def __init__(
             data.update(data[result_cols].fillna(value=0))
             data.loc[indices_with_null_val, "percent_expected_vote"] = 0
 
+        self.n_minimum_for_outlier_detection_model = 20
```
dmnapolitano, line 51: Doing this makes sense, but how did you come up with 20? 🤔

lennybronner: entirely arbitrary, but purposefully set higher than the 10 we need to run the model

jjcherian (approved):
Looks good to me!

PR #123, opened 2024-10-30
small changes to extrapolation

lennybronner:
## Description
I accidentally merged the earlier PR. This makes sure that the versioned dates work when we pass in strings, and also that logging works properly.


jjcherian (approved):
Looks good to me.

PR #124, opened 2024-10-31
Extrapolation rule fixes

lennybronner:
## Description
Just added some logging and also a small fix for when end_date of versioned results is None.

### DIFF
--- a/src/elexmodel/handlers/data/VersionedData.py
+++ b/src/elexmodel/handlers/data/VersionedData.py
@@ -11,6 +11,7 @@ from elexmodel.utils.file_utils import S3_FILE_PATH, TARGET_BUCKET
 
 LOG = logging.getLogger(__name__)
 
+
 class VersionedDataHandler:
     def __init__(
         self,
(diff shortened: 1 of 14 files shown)


dmnapolitano (commented):
Thanks for these changes! 🎉  I also noticed there's no mention of how to use the extrapolation rule in the README.  Can you add something? 🤔

lennybronner:
> Thanks for these changes! 🎉 I also noticed there's no mention of how to use the extrapolation rule in the README. Can you add something? 🤔

good call. readme updated!

dmnapolitano (approved):
Yay thank you!!! 🎉 🎉

src/elexmodel/handlers/data/VersionedData.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
dmnapolitano, line 1: Should this be `from elexmodel.logger import getModelLogger`? 🤔

lennybronner: yeah it should, good catch, thanks. unclear how that happened

PR #125, opened 2024-11-01
Remove min max electoral votes

lennybronner:
## Description
We can now remove the min/max at the bottom of electoral votes estimation, since that is properly handled by potential gains and losses. Also adds explicit dependencies for extrapolation.

## Test Steps
In the testbed, race calling as usual


dmnapolitano (approved):
LGTM!!!!! 🎉 🎉 🎉

PR #131, opened 2024-11-05
Error handling

jjcherian:
## Description

The data frame returned by versioned results is missing the est_correction column when there is an issue with the data.
This can result in a model failure if *every* unit is returned without an est_correction column in a state. "compute_correction_statistics" in BootstrapElectionModel.py will fail because df.est_correction.notnull() doesn't just return all null's, the attribute doesn't exist.


dmnapolitano (approved):
Thanks for finding this! 🎉

PR #138, opened 2024-11-06
hotfix for partial reporting bug

jjcherian:
## Description

truncation can fail catastrophically when % reporting is too low


dmnapolitano (approved):
Thank you!!! 🎉

PR #140, opened 2024-11-07
hotfix-extrap-house

jjcherian:
## Description

Hot fixes for the extrapolation step + using the presidential margins to infer a ticket splitting estimate in each house / senate race.

### DIFF
--- a/src/elexmodel/models/BootstrapElectionModel.py
+++ b/src/elexmodel/models/BootstrapElectionModel.py
@@ -52,7 +52,7 @@ class BootstrapElectionModel(BaseElectionModel):
     and the epsilons are contest (state/district) level random effects.
     """
 
-    def __init__(self, model_settings={}, versioned_data_handler=None):
+    def __init__(self, model_settings={}, versioned_data_handler=None, pres_predictions=None):
         super().__init__(model_settings)
         self.B = model_settings.get("B", 500)  # number of bootstrap samples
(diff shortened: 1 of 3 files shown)


src/elexmodel/models/BootstrapElectionModel.py
```diff
@@ -61,6 +61,10 @@ def __init__(self, model_settings={}, versioned_data_handler=None):
             "agg_model_hard_threshold", True
         )  # use sigmoid or hard thresold when calculating agg model
         self.district_election = model_settings.get("district_election", False)
+        self.office = model_settings.get("office", None)  # office of the election
+        self.election_id = model_settings.get("election_id", None)  # election ID
+        self.geographic_unit_type = model_settings.get("geographic_unit_type", None)  # geographic unit type
```
dmnapolitano: Where are these three fields used? 🤔

jjcherian: Oops dead code

dmnapolitano: No worries! 😄

dmnapolitano (approved):
Thanks, this is cool!  I wonder if, given this and your comment the other day about how our Senate baseline should be the same as the Presidential one, we should consider a hierarchical model in the future 👀

lennybronner (approved):

PR #145, opened 2024-11-20
ELEX-4647 more unit tests

dmnapolitano:
## Description

Hi!  The changes in this PR:

* ~~address `pylint` complaints, although `TODO` ones still remain;~~ now resolved in PR #146 
* ~~adds a `BaseDataHandler` abstract class that most `DataHandler`s are now subclasses of;~~ now resolved in PR #147
* adds some unit tests for `VersionedDataHandler` although I'm sure there's more scenarios we want to test there though I'm not exactly sure what 🤔 
* ensures that all top-level imports of external packages are dependencies in `setup.py`;

Thanks! 🎉 

## Jira Ticket

[ELEX-4647]

## Test Steps

`tox` or the `elexmodel` command of your choice 🎉 

[ELEX-4647]: https://arcpublishing.atlassian.net/browse/ELEX-4647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


setup.py
```diff
@@ -4,14 +4,15 @@
 from setuptools import find_packages, setup
 
 INSTALL_REQUIRES = (
+    "boto3>=1.34",
+    "botocore>=1.34.162",
     "click>=8.1",
+    "cvxpy~=1.5",
```
lennybronner, line 10: why are we adding this? Also curious why moving the other two?

src/elexmodel/handlers/s3.py
```diff
@@ -128,7 +132,7 @@ def wait_for_versions(self, q):
             try:
                 future.result()
                 yield version, data
-            except Exception as e:
+            except Exception as e:  # pylint: disable=broad-exception-caught
```
lennybronner: Don't we usually specify this with `# noqa` rather than pyline?

lennybronner (commented):
What's the purpose of adding the BaseDataHandler?

dmnapolitano:
> What's the purpose of adding the BaseDataHandler?

It's to unify the interface of all of our data handlers as much as possible and reduce some redundancy 🤷🏻‍♀️

setup.py
dmnapolitano: I noticed that `cvxpy` is a top-level import in `ConformalElectionModel.py`.  The moving of the other two is to put the requirements into alphabetical order 😄

src/elexmodel/handlers/s3.py
dmnapolitano: Hmmmmmm...I think you're right.  Let me see what happens if I switch all of these to `noqa` 🤔

lennybronner:
> > What's the purpose of adding the BaseDataHandler?
> 
> It's to unify the interface of all of our data handlers as much as possible and reduce some redundancy 🤷🏻‍♀️

Hmm, I see. I mean the LiveData and the PreprocessedData handler sort of makes sense. but the versioned data handler is very different. In this case the data is results rather then preprocessed data.

src/elexmodel/handlers/data/BaseDataHandler.py
```diff
@@ -0,0 +1,60 @@
...
+class BaseDataHandler(abc.ABC):
+    """
+    Abstract base handler for model data
+    """
+
+    def __init__(
+        self, election_id, office, geographic_unit_type, estimands, s3_client=None, historical=False, data=None
+    ):
+        self.election_id = election_id
+        self.office = office
```
lennybronner: I think this should be `office_id` rather than office, since that is how we reference it everywhere else

src/elexmodel/handlers/s3.py
dmnapolitano: Oh wow, so TIL that `pre-commit` looks at `noqa` lines which are part of the PEP8 standard, but `pylint` doesn't look at `noqa` at all 🤔  https://github.com/pylint-dev/pylint/issues/2493

So I guess I'll leave these `pylint` lines in so that `pylint` doesn't issue these complaints when we run `tox` 🤔

src/elexmodel/handlers/data/BaseDataHandler.py
dmnapolitano: Thanks, I agree!  I think I saw it as `self.office` in some of the data handlers and kept it lol 😄 🎉

dmnapolitano:
> > > What's the purpose of adding the BaseDataHandler?
> > 
> > 
> > It's to unify the interface of all of our data handlers as much as possible and reduce some redundancy 🤷🏻‍♀️
> 
> Hmm, I see. I mean the LiveData and the PreprocessedData handler sort of makes sense. but the versioned data handler is very different. In this case the data is results rather then preprocessed data.

Hmmmm, yeah, you're right.  There's a bit of overlap but not much and it doesn't really fit with the convention of preprocessed vs. results data.  Ok, I'll change it back now 😄

dmnapolitano:
> > > > What's the purpose of adding the BaseDataHandler?
> > > 
> > > 
> > > It's to unify the interface of all of our data handlers as much as possible and reduce some redundancy 🤷🏻‍♀️
> > 
> > 
> > Hmm, I see. I mean the LiveData and the PreprocessedData handler sort of makes sense. but the versioned data handler is very different. In this case the data is results rather then preprocessed data.
> 
> Hmmmm, yeah, you're right. There's a bit of overlap but not much and it doesn't really fit with the convention of preprocessed vs. results data. Ok, I'll change it back now 😄

Done! 🎉

src/elexmodel/handlers/s3.py
lennybronner: We didn't really deal with the issues with tox in the past, and I don't think we should moving forward. I think we can stick to just ones that make our pre-commit hooks fail.

src/elexmodel/handlers/data/VersionedData.py
```diff
@@ -143,15 +147,18 @@ def compute_estimated_margin(df):
             # Compute batch_margin using NumPy
             # this is the difference in dem_votes - the difference in gop_votes divided by the difference in total votes
             # that is, this is the normalized margin in the batch of votes recorded between versions
-            batch_margin = (
-                np.diff(results_dem, append=results_dem[-1]) - np.diff(results_gop, append=results_gop[-1])
-            ) / np.diff(results_weights, append=results_weights[-1])
+            with warnings.catch_warnings():
```
lennybronner: where is this coming from?

src/elexmodel/handlers/data/VersionedData.py
```diff
@@ -64,10 +65,12 @@ def get_versioned_results(self, filepath=None):
 
         if self.election_id.startswith("2020-11-03_USA_G"):
             path = "elex-models-prod/2020-general/results/pres/current.csv"
-        elif self.election_id.startswith("2024-11-05_USA_G"):
```
lennybronner: why has this been removed?

src/elexmodel/handlers/data/VersionedData.py
```diff
@@ -222,9 +231,8 @@ def get_versioned_predictions(self, filepath=None):
             return pd.read_csv(filepath)
 
         if self.election_id.startswith("2020-11-03_USA_G"):
-            path = "elex-models-prod/2020-general/prediction/pres/current.csv"
```
lennybronner: why has this been removed?

dmnapolitano: Once the exception is raised, `path` is unused.  We don't need it 🤷🏻‍♀️

src/elexmodel/handlers/data/VersionedData.py
dmnapolitano: It wasn't, the logic was a little redundant here so I cleaned it up right below this 😄

src/elexmodel/handlers/data/VersionedData.py
dmnapolitano: It happens if you have any zeros in the denominator.  Since we probably won't enable this until late into tab anyway, it should be pretty rare, but still, it's nice to not see it in our logs if we know it's a thing we're ok with 🤔

dmnapolitano: I just realized saying "it happens if you have zeros in the denominator" is not the most useful comment here lol 😅  More like, if you have any `results_weights` that are still zero due to units that haven't reported, especially in states with a lot of small units (New England, Texas) 🤔

lennybronner:
I am a little confused by these changes. It's not really what we had discussed. I would pretty strongly prefer to only include the unit tests in this PR (+ the changes to setup.py)

dmnapolitano:
> I am a little confused by these changes. It's not really what we had discussed. I would pretty strongly prefer to only include the unit tests in this PR (+ the changes to setup.py)

Hmmm ok, let me see what I can split into separate branches 🤔

dmnapolitano:
> > I am a little confused by these changes. It's not really what we had discussed. I would pretty strongly prefer to only include the unit tests in this PR (+ the changes to setup.py)
> 
> Hmmm ok, let me see what I can split into separate branches 🤔

Done! 🎉

lennybronner (approved):

PR #146, opened 2024-12-13
Resolve `pylint` and other warnings

dmnapolitano:
## Description

Hi!  The changes in this PR resolve warnings we see produced by `pylint` during `tox`, deprecation warnings from `pandas`, and other warnings produced by edge-cases that are expected.  Since warnings are often indicative of a potential problem, it's a good idea to address them so that it's easier to find new warnings encountered during future development work or other unexpected warnings generated during runtime.  Thanks! 🎉 

This work was previously part of #145 👍🏻 

## Test Steps

`tox`, testbed, `elexmodel` commands, etc. 🎉


src/elexmodel/handlers/data/VersionedData.py
```diff
@@ -143,15 +147,18 @@ def compute_estimated_margin(df):
             # Compute batch_margin using NumPy
             # this is the difference in dem_votes - the difference in gop_votes divided by the difference in total votes
             # that is, this is the normalized margin in the batch of votes recorded between versions
-            batch_margin = (
-                np.diff(results_dem, append=results_dem[-1]) - np.diff(results_gop, append=results_gop[-1])
-            ) / np.diff(results_weights, append=results_weights[-1])
+            with warnings.catch_warnings():
```
lennybronner, line 150: I vaguely remember us doing something previously to avoid this already. At least stopping it turning up in the logs. Am I misremembering?

src/elexmodel/handlers/s3.py
```diff
@@ -128,7 +132,7 @@ def wait_for_versions(self, q):
             try:
                 future.result()
                 yield version, data
-            except Exception as e:
+            except Exception as e:  # pylint: disable=broad-exception-caught
```
lennybronner, line 135: this is just to remove the tox error? We haven't really cared about those in the past, since we use our pre-commit hooks to make sure we are pep compliant.

tests/models/test_bootstrap_election_model.py
```diff
@@ -1063,7 +1063,7 @@ def test_get_national_summary_estimates(bootstrap_election_model, rng):
     bootstrap_election_model.get_aggregate_predictions(
         reporting_units, nonreporting_units, unexpected_units, ["postal_code"], "margin"
     )  # race calling for aggregate prediction interval assumes that the point prediction has been set accordingly
-    lower, upper = bootstrap_election_model.get_aggregate_prediction_intervals(
+    _, _ = bootstrap_election_model.get_aggregate_prediction_intervals(
```
lennybronner, line 1066: I could have sworn I already did this at some point

src/elexmodel/handlers/data/VersionedData.py
dmnapolitano: Interesting 🤔  Maybe it was another spot where we can occasionally get divisions by zero? Without this, right now I get warnings here when running the unit tests 🤔  Curious 🤔

src/elexmodel/handlers/s3.py
dmnapolitano: In that case maybe we should disable `pylint` altogether.  If we're just going to ignore its complaints, we might as well speed up our `tox` runs 😄 

Another option here would be to add more things to our list of `pylint` complaints we ignore in `setup.cfg`, although I'm hesitant to do that with too many things because a lot of these are useful, but there are cases where we want to ignore them.  This one in particular is here because `boto3` and its related libraries can raise several different types of `Exception`s and we want to handle them all the same way.  Ideally, we'd break that out or specify the types but honestly, I find it hard to know up front which exception types `boto3` can throw 😬

tests/models/test_bootstrap_election_model.py
dmnapolitano: This is the PR of déjà vu 😂 

If it helps, it looks like you made similar changes in PR #109 😄 👍🏻

src/elexmodel/handlers/s3.py
lennybronner: Is there any downside to disabling pylint? Do we use it in other live team repos? If we do, I think we just ignore the complaints.

src/elexmodel/handlers/data/VersionedData.py
lennybronner: If you re-run the 2024 in the testbed (without writing anything to s3), do you see this warning?

src/elexmodel/handlers/s3.py
dmnapolitano: I'll ask, but if they're using `pylint`, they're probably not ignoring the complaints.  In which case, I'm not sure why we would either.  Sure, some of these are annoying, but we can put in rules to have `pylint` not issue annoying complaints about benign things (e.g. "too many arguments") and keep the ones that are useful and point towards potential problems 👀

src/elexmodel/handlers/data/VersionedData.py
dmnapolitano: Alright so I ran this:

```
python run.py 2024-11-05_USA_G redo --office_id S --estimands="['margin']" --features "['baseline_normalized_margin']" --end_timestamp 2024-11-06T12:00:00-05:00 --model_parameters='{"extrapolation" : True, "versioned_start_date" : "2024-11-05T21:00:00-05:00", "versioned_end_date" : "2024-11-05T21:30:00-05:00"}
```

And the answer right now is no, BUT, I noticed that even though we're capturing warnings, they're not currently coming through in the logs.  Having changed that, this actually generates a lot of warnings 😬  

PR incoming 👀

src/elexmodel/handlers/s3.py
lennybronner: I think the answer is we should stop using tox and use pytest directly in the unit tests. Since there is no need to use two different linters

dmnapolitano: Hmmmmmmm...I'd argue in favor of using `tox` but removing `pylint` from it.  `tox` is great because we can specify multiple Python versions and it'll run whatever suite of tests etc. for each version in isolated environments. We can also have it run `pre-commit` for us, e.g. https://github.com/jugmac00/flask-reuploaded/blob/master/tox.ini 🤔   But yeah it looks pretty easy to remove `pylint` from `tox`, which is good 😄

dmnapolitano: @lennybronner alright so I did some reading here and it seems like `pylint` is designed with the intention of being overly-cautious.  The best way I saw it described is that `pylint` is a superset of `flake8`.  We probably don't need the extra complaints `pylint` generates on top of `flake8` 🤔 

So yeah, let me know if you want me to remove `pylint` in this PR or open another one 😄

src/elexmodel/handlers/data/VersionedData.py
dmnapolitano: Hi @jjcherian ! Hope you're doing well 😄 👋🏻 

We noticed that these few lines can generate quite a few `invalid value encountered in divide` and `divide by zero encountered in divide` warnings.  If you clone the `develop` branch and run, for example,

```
elexmodel 2024-11-05_USA_G --estimands=margin --features=baseline_normalized_margin --office_id=S --geographic_unit_type=county --pi_method bootstrap --national_summary --model_parameters '{"extrapolation" : True, "versioned_start_date" : "2024-11-05T21:00:00-05:00", "versioned_end_date" : "2024-11-05T21:30:00-05:00"}'
```

you should see them.  We were wondering what the best way to handle these warning is. Right now, in this PR, I've added code to ignore (and therefore silence) them, but do you think there's a better way to resolve these? 🤔 

Thanks in advance 😄

